PR #17, opened 2022-06-10
fix: fix sign in transaltions

subbotnik16:


victor-oancea-leftclick-io:
don't merge it, I changed that part

PR #27, opened 2022-06-24
Settings screen

pisarevkir:


src/navigation/components/Header/components/LangButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <View style={[styles.container, containerStyle]}>
+        <Text style={[styles.langText, {color}]}>EN</Text>
+        <WorldIcon fill={color} style={styles.worldIcon} />
+        <DropdownIcon fill={color} style={styles.dropdownIcon} />
+      </View>
+    </TouchableOpacity>
+  );
+};
+
+const buttonHitSlop = {top: 4, left: 4, bottom: 4, right: 4};
```
OlehPanforov, line 45: Let's move buttons hit slop to constants, I see you are using it in the SettingsButton also.
Maybe we can even make some sort of increased hit slop in global constants and just apply it in any place when needed (when the button is less than 44px width/height for example)

victor-oancea-leftclick-io (approved):

PR #49, opened 2022-08-08
Profile image

pisarevkir:


src/screens/SettingsFlow/PersonalInformation/index.tsx
```diff
@@ -97,6 +102,15 @@ export const PersonalInformation = memo(() => {
           <Animated.View style={[styles.avatar, animatedAvatarStyle]}>
             <Avatar
               showPen
+              onPenPress={() => {
+                ImagePicker.openPicker({
+                  width: 400,
```
victor-oancea-leftclick-io, line 107: Free flow, not fixed

victor-oancea-leftclick-io (changes requested):

src/components/Avatar/index.tsx
```diff
@@ -18,17 +19,29 @@ const AVATAR_SIZE = rem(86);
...
+  style,
+  showPen,
+  onPenPress,
+}: Props) => {
   return (
     <View style={[style, isIOS && commonStyles.shadow]}>
       <View style={[styles.avatarWrapper, isAndroid && commonStyles.shadow]}>
-        <Image source={{uri}} style={styles.image} />
+        {uri &&
+          (useLocalImage ? (
```
victor-oancea-leftclick-io, line 40: why?

src/screens/ProfileFlow/Profile/components/UserInfo/components/Avatar.tsx
```diff
@@ -1,33 +1,25 @@
...
 
 type Props = {
-  uri: string;
-  showPen: boolean;
-  onPenPress?: () => void;
+  uri?: string | null;
 };
 
-export const Avatar = ({uri, showPen, onPenPress}: Props) => {
+export const Avatar = ({uri}: Props) => {
```
victor-oancea-leftclick-io, line 15: Why not use the global one?

src/screens/Home/components/Header/index.tsx
```diff
@@ -12,16 +13,22 @@ import {StyleSheet, Text, TouchableOpacity, View} from 'react-native';
...
 
-export const HomeHeader = ({}: HomeHeaderProps) => {
+export const HomeHeader = () => {
   const user = useSelector(userSelector);
   return (
     <View style={styles.container}>
       <View style={styles.info}>
-        <View style={styles.image} />
+        {user?.profilePictureUrl && (
+          <RemoteImage
```
victor-oancea-leftclick-io, line 24: Use the avatar component

src/screens/SettingsFlow/PersonalInformation/hooks/usePickImage.ts
```diff
@@ -0,0 +1,22 @@
...
+export const usePickImage = ({
+  onImageSelected,
+  options,
+}: {
+  onImageSelected: (image: Image) => void;
+  options?: Options;
+}) => {
+  const openPicker = () =>
+    ImagePicker.openPicker({
+      width: 400,
```
victor-oancea-leftclick-io, line 14: is that the default one or is it fixed?

src/screens/SettingsFlow/PersonalInformation/index.tsx
```diff
@@ -97,7 +116,9 @@ export const PersonalInformation = memo(() => {
           <Animated.View style={[styles.avatar, animatedAvatarStyle]}>
             <Avatar
               showPen
-              uri="https://media.istockphoto.com/photos/millennial-male-team-leader-organize-virtual-workshop-with-employees-picture-id1300972574?b=1&k=20&m=1300972574&s=170667a&w=0&h=2nBGC7tr0kWIU8zRQ3dMg-C5JLo9H2sNUuDjQ5mlYfo="
+              onPenPress={openPicker}
+              uri={userDraft.profilePicture?.uri ?? user.profilePictureUrl}
+              useLocalImage={Boolean(userDraft.profilePicture?.uri)}
```
victor-oancea-leftclick-io, line 121: why do you need this exposed? This should be the internal business of that component. It should have a "dirty" state when you select any photo, which it should clean (re-feed from uri) on render

victor-oancea-leftclick-io (changes requested):
x

src/screens/SettingsFlow/PersonalInformation/hooks/usePickImage.ts
pisarevkir: You mean the width / height params? This is the size of an image in px after cropping. And those numbers affect the shape of the crop shape (width = 400, height = 400 -> square)

victor-oancea-leftclick-io: can I increase/decrease that cropping 400 px square?

src/screens/SettingsFlow/PersonalInformation/index.tsx
pisarevkir: Unfortunately the component should distinguish between uri from the BE (where we have to add width as query param) and the uri from the local file system, cuz if we add query params to the local uri it stops working. Actually i can take out the query params adding to the util function and control the uri outside of the Avatar component..

src/components/Avatar/index.tsx
pisarevkir: https://github.com/ice-blockchain/mobile-app/pull/49#discussion_r940267866

src/screens/SettingsFlow/PersonalInformation/index.tsx
victor-oancea-leftclick-io: thats not the question im asking. Ofc you have 2 states. My question is : why do you need to expose that info to the users of that component? Just make it hidden. On the callback from onPenPress you set a "dirty" state inside the component, which you clear on the next rendering

src/screens/ProfileFlow/Profile/components/UserInfo/components/Avatar.tsx
pisarevkir: This one has different styles + no edit button or something, so using global one has no benefits

src/screens/SettingsFlow/PersonalInformation/index.tsx
pisarevkir: What do you mean

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
```diff
@@ -1,38 +1,36 @@
...
 type Props = {
   focused: boolean;
 };
 
-const PROFILE_IMAGE_SIZE = 22;
-const DEV_PROFILE_IMAGE =
-  'https://static.independent.co.uk/s3fs-public/thumbnails/image/2016/08/26/13/mr-bean.jpg';
+const PROFILE_IMAGE_SIZE = rem(22);
 
 export const ProfileIcon = ({focused}: Props) => {
```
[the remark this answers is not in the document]
pisarevkir: This one has different styles + no edit button or something, so using global one has no benefits

src/screens/SettingsFlow/PersonalInformation/hooks/usePickImage.ts
pisarevkir: Yea, you can zoom the image during the cropping

src/screens/SettingsFlow/PersonalInformation/index.tsx
pisarevkir: How do the component should understand to use the simple `Image` component (without adding query params) or RemoteImage to display the image having only uri?

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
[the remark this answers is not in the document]
victor-oancea-leftclick-io: put the styles as args, and the edit button can be missing. So it has all the benefits in the world to reuse something

src/screens/SettingsFlow/PersonalInformation/index.tsx
victor-oancea-leftclick-io: if its dirty or not

victor-oancea-leftclick-io: instead of requiring that useLocalImage bool, have an private/internal dirty flag that you change after user picks an image. dirty ? <Image/> : <RemoteImage/>

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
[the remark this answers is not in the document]
pisarevkir: If we pass styles as props, then the Avatar component don't bring any benefits - under the hood that is only a wrapper on RemoteImage with some styles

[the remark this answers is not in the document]
pisarevkir: i'll rethink the Avatar component

src/screens/ProfileFlow/Profile/components/UserInfo/components/Avatar.tsx
pisarevkir: i'll rethink the Avatar component

src/components/Avatar/index.tsx
pisarevkir: now internal Avatar component state controls what to show - uri from props or local state image

src/screens/ProfileFlow/Profile/components/UserInfo/components/Avatar.tsx
pisarevkir: Here the global one is used

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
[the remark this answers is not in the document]
pisarevkir: I still think it's better to use plain RemoteImage here - it doesn't make any sense to  use Avatar component that has a lot of additional logic inside that is used only when editing is allowed (useState, usePickImage, useEffect, several additional Views - that all affect performance and doesn't bring any benefits in this case).

src/screens/Home/components/Header/index.tsx
pisarevkir: I still think it's better to use plain RemoteImage here - it doesn't make any sense to  use Avatar component that has a lot of additional logic inside that is used only when editing is allowed (useState, usePickImage, useEffect, several additional Views - that all affect performance and doesn't bring any benefits in this case).

src/screens/SettingsFlow/PersonalInformation/index.tsx
pisarevkir: now internal Avatar component state controls what to show - uri from props or local state image, this prop is not exposed anymore

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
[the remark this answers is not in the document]
pisarevkir: This is the same case as when we display the list of users - we use simple RemoteImage there, but not the Avatar component - cuz the editing is never the case and the ListUserItem has to be as simple as possible

src/components/Avatar/index.tsx
```diff
@@ -1,45 +1,91 @@
...
-    <View style={[style, isIOS && commonStyles.shadow]}>
-      <View style={[styles.avatarWrapper, isAndroid && commonStyles.shadow]}>
-        <Image source={{uri}} style={styles.image} />
+export const Avatar = memo(
+  ({
+    uri,
+    editable,
+    onChange,
+    avatarSize = DEFAULT_AVATAR_SIZE,
+    containerStyle,
```
victor-oancea-leftclick-io, line 40: u can set defaults for all 3 styles

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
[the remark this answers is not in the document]
victor-oancea-leftclick-io: then do this: create 2 avatars: what you have atm and ReadOnlyAvatar, and use only those 2, no more custom profile pic everywhere.

Keep edit logic in avatar and move the rest in ReadOnlyAvatar

src/screens/Home/components/Header/index.tsx
victor-oancea-leftclick-io: same reply

victor-oancea-leftclick-io (changes requested):

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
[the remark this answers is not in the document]
pisarevkir: This would be reasonable if we had some custom logic / styles / whatever inside, but it would be just
```
export const ReadOnlyAvatar = RemoteImage;
```

Cuz basically `ReadOnlyAvatar` is just a RemoteImage (same props: uri, style and width) -> nothing to incapsulate.

src/components/Avatar/index.tsx
pisarevkir: those are optional additional styles:
containerStyle - is used for element positioning and has no default value
imageWrapperStyle - is used if you want to overwrite some of the styles.avatarWrapper
imageStyle - is used if you want to change the some of the styles.image styles

If we set `styles.avatarWrapper` as default for `imageWrapperStyle` prop, then if we pass that prop to component, we'll have to set all the styles for that view. But the way it works now, you can overwrite some specific style property without flushing the rest.

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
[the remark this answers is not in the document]
pisarevkir: now the Avatar component is used

src/screens/Home/components/Header/index.tsx
pisarevkir: now the Avatar component is used

src/components/Avatar/EditableAvatar.tsx
```diff
@@ -0,0 +1,110 @@
...
+} from 'react-native';
+import {Image as CropImage} from 'react-native-image-crop-picker';
+import {rem} from 'rn-units';
+
+const DEFAULT_AVATAR_SIZE = rem(86);
+const PEN_SIZE = rem(22);
+
+type Props = {
+  uri?: string;
+  editable?: boolean;
```
victor-oancea-leftclick-io, line 25: Whats the point of this? The whole component states that is is only editable

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
```diff
@@ -1,37 +1,29 @@
...
-          <Image source={{uri: DEV_PROFILE_IMAGE}} style={styles.image} />
-        </View>
+    <View
+      style={[styles.imageBorder, focused ? styles.imageBorderFocused : null]}>
+      {user?.profilePictureUrl && (
+        <Avatar
+          uri={user.profilePictureUrl}
+          size={rem(22)}
+          borderRadius={rem(11)}
+          allowFullScreen={false}
```
victor-oancea-leftclick-io, line 25: Why not?

src/screens/SettingsFlow/PersonalInformation/index.tsx
```diff
@@ -94,10 +112,17 @@ export const PersonalInformation = memo(() => {
...
-              uri="https://media.istockphoto.com/photos/millennial-male-team-leader-organize-virtual-workshop-with-employees-picture-id1300972574?b=1&k=20&m=1300972574&s=170667a&w=0&h=2nBGC7tr0kWIU8zRQ3dMg-C5JLo9H2sNUuDjQ5mlYfo="
+          <Animated.View
+            style={[
+              styles.avatarWrapper,
+              commonStyles.shadow,
+              animatedAvatarStyle,
+            ]}>
+            <EditableAvatar
+              uri={user.profilePictureUrl}
+              editable={true}
```
victor-oancea-leftclick-io, line 123: Redundant

src/utils/openSms.ts
```diff
@@ -1,12 +0,0 @@
-// SPDX-License-Identifier: BUSL-1.1
```
victor-oancea-leftclick-io, line 1: Why did u delete it?

src/screens/ImageView/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const {
+    params: {imageRef, uri, size, borderRadius = 0},
+  } = useRoute<RouteProp<MainStackParamList, 'ImageView'>>();
+
+  const {top: topInset, bottom: bottomInset} = useSafeAreaInsets();
+  const viewPortWidth = screenWidth;
+  const viewPortHeight = screenHeight - topInset - bottomInset - HEADER_HEIGHT;
+
+  const {imageMeasure} = useImageMeasure(imageRef);
+  const {imageUri} = useImageSizedUri(uri, size);
```
victor-oancea-leftclick-io, line 28: Where are you getting the 1920x1920 picture?

victor-oancea-leftclick-io (changes requested):

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+type Props = {
+  uri?: string;
+  size?: number;
+  borderRadius?: number;
+  allowFullScreen?: boolean;
+  style?: StyleProp<ImageStyle>;
+};
+
+export const Avatar = memo(
```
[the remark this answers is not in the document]
pisarevkir: This is a perf optimization - the component will not be rerendered if the props are not changed

src/components/Avatar/EditableAvatar.tsx
pisarevkir: removed

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
pisarevkir: In tabbar we open the corresponding tab on press instead of the full screen image

src/screens/SettingsFlow/PersonalInformation/index.tsx
pisarevkir: removed

src/utils/openSms.ts
pisarevkir: the idea is to group util functions by domain (`utils/file` `utils/network` and so on), so i put that function to the `utils/device` file

src/screens/ImageView/index.tsx
pisarevkir: `useImageSizedUri` returns uri for a full screen image as soon as it is ready, so the flow if like so:
* `useImageSizedUri` returns uri for the small image and starts loading the image with the full screen width
* as soon as the full width image loaded (if it's already in cache, it will be returned right away), the uri is changed to the new one

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
victor-oancea-leftclick-io: What if the image is updated from BE, do you display that or the cached one?

src/navigation/components/MainTabBar/components/Icons/ProfileIcon.tsx
victor-oancea-leftclick-io: Right

src/utils/openSms.ts
victor-oancea-leftclick-io: Great

src/screens/ImageView/hooks/useImageSizedUri.ts
```diff
@@ -0,0 +1,19 @@
...
+import {Image} from 'react-native';
+import {screenWidth} from 'rn-units';
+
+export const useImageSizedUri = (uri: string, initialSize: number) => {
+  const [imageUri, setImageUri] = useState(
+    getImageUriForSize(uri, {width: initialSize}),
+  );
+
+  useEffect(() => {
+    const fullScreeUri = getImageUriForSize(uri, {width: screenWidth});
```
victor-oancea-leftclick-io, line 14: You’ll lose quality if u do that. Just put 1920. Cuz if you’re using low res phone, lets say 480, and you pinch out, you have only 480 pixels that get enlarged, rather than having the whole 1920 pixels to enjoy

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
pisarevkir: If the image is updated -> the `uri` is changed -> the component will get updated. `memo` caches the component only if the props stay the same

victor-oancea-leftclick-io (dismissed):

victor-oancea-leftclick-io (approved):

PR #56, opened 2022-08-25
Tabbar redesign

ice-orion:


ice-hector (dismissed):

src/store/modules/Economy/reducer/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  return produce(state, draft => {
+    switch (action.type) {
+      case EconomyActions.STORE_MINIG_TOOLTIP_SEEN.STATE.type:
+        draft.isMiningTooltipSeen = true;
+        break;
+    }
+  });
+}
+
+const persistConfig = {
```
ice-ares, line 30: U dont need to do this, u'll get it from the BE. Its the "active" field, if its true, mining is in progress

robertpreoteasa (dismissed):

ice-eros (approved):

src/store/modules/Economy/reducer/index.ts
ice-orion: You're about the flag `isMiningTooltipSeen`? It means that the tooltip was displayed (and this flag should be persisted, so when a user reopens the app / relogins the tooltip is not being displayed anymore). And we can't rely on mining.active -> mining could be not active at some point but the tooltip still has to be hidden.

PR #80, opened 2022-11-29
Feature/issues fixing

ice-hector:


src/screens/WelcomeFlow/SetEmail/hooks/useSetEmail.ts
```diff
@@ -28,8 +28,10 @@ export const useSetEmail = () => {
 
   const onSubmit = () => {
     Keyboard.dismiss();
-    dispatch(ValidationActions.EMAIL_VALIDATION.RESET.create());
-    dispatch(AccountActions.UPDATE_ACCOUNT.START.create({email}));
+    if (email !== '') {
+      dispatch(ValidationActions.EMAIL_VALIDATION.RESET.create());
+      dispatch(AccountActions.UPDATE_ACCOUNT.START.create({email}));
```
ice-orion, line 33: Usually it's not good to block the UI without notifying user about the problem. Moreover email validation already exists and happens in `updateAccountSaga`

```
if ('email' in userInfo && !validateEmail(userInfo.email ?? '')) {
      throw new Error(t('errors.invalid_email'));
}
```

You can add a separate error message for the empty email if needed (now `Invalid email` is throwed)

src/navigation/components/MainTabBar/index.tsx
```diff
@@ -77,5 +77,6 @@ const styles = StyleSheet.create({
   buttons: {
     flexDirection: 'row',
     borderBottomColor: COLORS.white,
+    backgroundColor: COLORS.white,
```
ice-orion, line 80: It breaks the UI, this view background should be transparent

src/screens/WelcomeFlow/SetEmail/hooks/useSetEmail.ts
ice-hector: Ok will add, because right now no checks for the empty email and we send a request to server with empty one

src/navigation/components/MainTabBar/index.tsx
ice-orion: The shadow becomes non-transparent

ice-hector: Yeah. I will ask designer to add the white bg there

src/store/modules/Account/sagas/signOut.ts
```diff
@@ -1,15 +1,26 @@
...
+import {getErrorMessage} from '@utils/errors';
 import {call, put} from 'redux-saga/effects';
 
 export function* signOutSaga() {
   try {
     yield call(signOut);
     yield put(AccountActions.SIGN_OUT.SUCCESS.create());
   } catch (error) {
+    const localizedError = getErrorMessage(error);
+    navigationRef.navigate('ErrorPopUp', {
```
ice-ares, line 16: can u move this logic in that utils/errors? so we can just call it as `showError(error)`

ice-hector: Yeah sure

ice-hector: Fixed

ice-ares (dismissed):

ice-hades (dismissed):

ice-orion (dismissed):

ice-ares (dismissed):

src/navigation/Main.tsx
```diff
@@ -140,8 +140,8 @@ export type TeamTabStackParamList = {
 
 export type ProfileTabStackParamList = {
   Profile: {userId: string} | undefined;
```
ice-orion, line 142: This is the Profile screen for my-profile, isn't it? Probably it makes sense to rename it to `MyProfile` then? And it seems we don't need to pass `userId` to the `Profile`, `MyRoles`, `MyBadges` screens since those screens are always for the current user

src/navigation/Main.tsx
```diff
@@ -118,16 +118,16 @@ export type MainStackParamList = {
   CountrySelect: {
     onSelect: (country: Country) => void;
   };
+  UserProfile: {userId: string} | undefined;
+  MyRoles: {userId?: string} | undefined;
+  MyBadges: {category?: BadgeCategory; userId?: string};
```
ice-orion, line 123: May be it would be better to rename `MyRoles` -> `UserRoles`, `MyBadges` -> `UserBadges`?

src/screens/ProfileFlow/Profile/components/AvatarHeader/index.tsx
```diff
@@ -136,17 +137,23 @@ export const AvatarHeader = memo(
...
 
     return (
       <Animated.View style={[styles.container, extraPadding, shadowStyle]}>
-        <BackButton
-          containerStyle={styles.backButton}
-          color={COLORS.primaryDark}
-          allowOnTab
-        />
+        <View style={styles.leftContainer}>
+          {!isOwner && (
```
ice-orion, line 149: Hypothetically we can end up opening our own user in the FullScreen page (from a push / deeplink / if we can find our own user in some list) and in this case we'll stuck in the full-screen page without possibility to go back. If the idea is no hide the back button for the tab but show for the stack, then just remove `allowOnTab` prop from the `BackButton`

src/screens/ProfileFlow/Profile/index.tsx
```diff
@@ -30,29 +30,38 @@ import Animated, {
...
 
 export const Profile = memo(() => {
   const [isTooltipVisible, setIsTooltipVisible] = useState(false);
   const [contactDetails, setContactDetails] = useState<Contact>();
+  const authUser = useSelector(userSelector);
+  const route = useRoute<RouteProp<ProfileTabStackParamList, 'Profile'>>();
+  const isOwner = !route.params || route.params.userId === authUser?.id;
+
+  const bottomOffset = useBottomTabBarOffsetStyle({
+    extraOffset: isOwner ? rem(64) : 0,
```
ice-orion, line 45: no need to rely on `isOwner` in this case - can't we just use it without overwriting `extraOffset`?
```
const bottomOffset = useBottomTabBarOffsetStyle()
```
?

ice-hector: Yeah sorry, this was before I refactored TabBarOffset hook

src/navigation/Main.tsx
ice-hector: Renamed to Roles and Badges

src/navigation/Main.tsx
ice-hector: No, MyRoles and MyBadges are available for other users also, that's why we need userId property. 
Profile in the tab will rename to MyProfile because only own profile can be opened there, yep.

ice-hector: And will remove userId param from MyProfile screen

ice-orion: > No, MyRoles and MyBadges are available for other users also
Yea, but we have the `Roles` and `Badges` screens in `MainStack` because of that, and those screens accept `userId` because they are for other users, but the screens in ProfileTab are always for my user, aren't they?

src/screens/ProfileFlow/Profile/components/AvatarHeader/index.tsx
ice-hector: Ok I removed allowOnTab, now back button still hidden on tab and never hidden when presented full screen

ice-orion (approved):

ice-ares (approved):

PR #83, opened 2022-12-02
chore: stats/user growth API integration

ice-hades:


src/store/modules/Stats/selectors/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import {TimeSeriesStats} from '@store/modules/Stats/reducer';
+import {StatsPeriod} from '@store/modules/Stats/types';
+import {RootState} from '@store/rootReducer';
+
+export const getStatsSelector =
+  (period: StatsPeriod) =>
+  (state: RootState): TimeSeriesStats => {
+    switch (period) {
+      case 3:
```
ice-ares, line 11: u could create a map of period:stats

src/screens/HomeFlow/Stats/components/UsersGrowthGraph/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function getEmptyData(statsPeriod: StatsPeriod) {
+  return Array(statsPeriod)
+    .fill(null)
+    .map((_, i) => {
+      const date = new Date();
+      date.setDate(date.getDate() - i);
+      return {
+        label: `${date.getDate()}/${date.getMonth()}`,
```
ice-ares, line 23: use the US format, with MM/DD

src/screens/HomeFlow/Stats/components/UsersGrowthGraph/utils.ts
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {TimeSeries} from '@api/statistics/types';
+import {BarGraphData} from '@components/BarGraph/types';
+import {StatsPeriod} from '@store/modules/Stats/types';
+
+export function getLabel(timeSeries: TimeSeries) {
+  const date = timeSeries.date.split('T'); //2022-12-01T11:52:35.348Z
```
ice-ares, line 8: use dayjs for working with dates

ice-ares (changes requested):

src/screens/HomeFlow/Stats/components/UsersGrowthGraph/utils.ts
ice-hades: That was my first approach. But there are issues because of timezones. after translating the date from BE to Date JS dates there can be dates with the same day/month pairs

ice-ares: that's not possible, the dates are UTC and the intervals are unique, never overlapping

ice-hades: Here is the response from Swagger for 3 days period
`{
  "timeSeries": [
    {
      "date": "2022-12-02T16:40:23.510731588Z",
      "active": 0,
      "total": 1
    },
    {
      "date": "2022-12-01T23:59:59.999999999Z",
      "active": 0,
      "total": 0
    },
    {
      "date": "2022-11-30T23:59:59.999999999Z",
      "active": 0,
      "total": 0
    }
  ],
  "active": 0,
  "total": 1
}`

ice-ares: Cant happen cuz u shift all of them with the timezone

ice-ares: Actually, u’re right. Handle the first element differently. Round it up

ice-hades: What do you mean round up? When I convert it to Date JS it is already can be a previous day because of the timezone shift.

ice-ares: Add the nanos to the first date until u get to 23:59.999999999 UTC

ice-hades: time doesn't matter already. when you create a Date JS from a string it is already converted to your timezone and you already can have 2 dates with the same day/month pairs. Doesn't matter if you set time to 23:59.999999999

ice-hades: If you so much insists it is possible with DateJS you can suggest a working solution. Here is the test case. You've got 2 strings "2022-12-01T23:59:59.999999999Z" and "2022-12-02T16:40:23.510731588Z". COnvert them to Date JS so that getDate() would return different values

ice-ares: All in saying is u can use dayjs for working with dates instead of parsing and spliting strings in a hardcoded way

ice-hades: @ice-ares please suggest then a solution where given "2022-12-01T23:59:59.999999999Z" and "2022-12-02T16:40:23.510731588Z" strings you convert them to Date JS so that getDate() would return different values

ice-ares: cant help you with that. Maybe @ice-orion can

src/store/modules/Stats/actions/index.ts
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {UserGrowth} from '@api/statistics/types';
+import {StatsPeriod} from '@store/modules/Stats/types';
+import {createAction} from '@store/utils/actions/createAction';
+
+const GET_THREE_DAYS_STATS = createAction('GET_THREE_DAYS_STATS', {
```
ice-orion, line 7: Let's create one actions that takes period as param? If you need to track loading / errors separately for every provided period you can utilise `isMultiInstanceProcess: true` flag (check `NEWS_POST_LOAD` action that separates that data for every news post)

src/store/modules/Stats/actions/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  START: () => ({}),
+  SUCCESS: (userGrowth: UserGrowth) => ({userGrowth}),
+  FAILED: () => ({}),
```
ice-orion, line 10:
```
START: true,
SUCCESS: (userGrowth: UserGrowth) => ({userGrowth}),
FAILED: (errorMessage: string) => ({errorMessage}),
```

src/store/modules/Stats/reducer/index.ts
```diff
@@ -0,0 +1,96 @@
...
+import {TimeSeries} from '@api/statistics/types';
+import {AccountActions} from '@store/modules/Stats/actions';
+import produce from 'immer';
+
+export interface TimeSeriesStats {
+  status: 'PENDING' | 'LOADING' | 'FAILED' | 'LOADED';
+  timeSeries: TimeSeries[];
+}
+
+export interface StatsState {
```
ice-orion, line 12: We can simplify the store like this - 
```
export interface StatsState {
  active: number;
  total: number;
  timeSeries: {[number: StatsPeriod]: TimeSeries[] | null}
}
```

So as i described above, we don't store `status` manually + since we can have only one action that also has `period` as param it could be much easier to store the results in `{period: data}` way

src/screens/HomeFlow/Stats/components/UsersGrowthGraph/utils.ts
ice-orion: `dayjs.utc('2022-12-01T23:59:59.999999999Z').date() -> 1`
`dayjs.utc('2022-12-02T16:40:23.510731588Z').date() -> 2`

src/screens/HomeFlow/Stats/components/Summary/index.tsx
```diff
@@ -1,23 +1,32 @@
...
   return (
     <View style={styles.container}>
       <View style={[styles.cell, styles.cell_left]}>
         <LogoIcon color={COLORS.white} width={rem(32)} height={rem(32)} />
         <View style={styles.cellDataLeft}>
           <Text style={styles.labelText}>{t('stats.online_miners')}</Text>
           <View style={styles.value}>
-            <Text style={styles.valueText}>25,141</Text>
+            <Text style={styles.valueText}>
+              {totalActiveUsers.toLocaleString('en-US')}
```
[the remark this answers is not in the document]
ice-hades: https://reactnative.dev/blog/2022/07/08/hermes-as-the-default#ios-intl

src/store/modules/Stats/selectors/index.ts
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {TimeSeriesStats} from '@store/modules/Stats/reducer';
+import {StatsPeriod} from '@store/modules/Stats/types';
+import {RootState} from '@store/rootReducer';
+
+export const getStatsSelector =
```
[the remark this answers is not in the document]
ice-hades: you see it is not a selector itself. it returns a selector

src/screens/HomeFlow/UserGrowthGraph/index.tsx
```diff
@@ -10,19 +10,22 @@ import {useBottomTabBarOffsetStyle} from '@navigation/hooks/useBottomTabBarOffse
 import {useFocusStatusBar} from '@navigation/hooks/useFocusStatusBar';
 import {HomeTabStackParamList} from '@navigation/Main';
 import {RouteProp, useRoute} from '@react-navigation/native';
-import {
-  MOCK_ACTIVE_USERS_GRAPH_DATA,
-  MOCK_TOTAL_USERS_GRAPH_DATA,
-} from '@screens/HomeFlow/Stats/components/UsersGrowthGraph/mockData';
+import {useGetBarGraphDataForStatsPeriod} from '@screens/HomeFlow/Stats/components/UsersGrowthGraph/hooks/useGetBarGraphDataForStatsPeriod';
```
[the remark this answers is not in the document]
ice-hades: MOCK_TOTAL_USERS_GRAPH_DATA is still used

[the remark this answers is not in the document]
ice-hades: Sorry, misplaced the response

src/screens/HomeFlow/Stats/components/UsersGrowthGraph/mockData.ts
```diff
@@ -8,12 +8,3 @@ export const MOCK_TOTAL_USERS_GRAPH_DATA = Array(30)
     value: Math.round(Math.random() * totalMax),
   }))
   .reverse();
-
```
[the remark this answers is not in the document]
ice-hades: MOCK_TOTAL_USERS_GRAPH_DATA is still used

src/store/modules/Stats/actions/index.ts
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {UserGrowth} from '@api/statistics/types';
+import {StatsPeriod} from '@store/modules/Stats/types';
+import {createAction} from '@store/utils/actions/createAction';
+
+const GET_PERIOD_STATS = createAction('GET_PERIOD_STATS', {
```
ice-ares, line 7: get user growth stats

src/store/modules/Stats/hooks/useGetBarGraphDataForStatsPeriod.ts
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {AccountActions} from '@store/modules/Stats/actions';
+import {getUsersBarGraphDataSelector} from '@store/modules/Stats/selectors';
+import {StatsPeriod, UsersBarGraphData} from '@store/modules/Stats/types';
+import {useEffect} from 'react';
+import {useDispatch, useSelector} from 'react-redux';
+
+export function useGetBarGraphDataForStatsPeriod(
```
ice-ares, line 9: userGrowthStats, not just stats

src/store/modules/Stats/reducer/index.ts
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {TimeSeries} from '@api/statistics/types';
+import {AccountActions} from '@store/modules/Stats/actions';
+import produce from 'immer';
+
+export interface StatsState {
```
ice-ares, line 7: this is not a generic stats state, this is specifically only for user growth

src/store/modules/Stats/sagas/getPeriodStats.ts
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {Api} from '@api/index';
+import {UserGrowth} from '@api/statistics/types';
+import {AccountActions} from '@store/modules/Stats/actions';
+import {getErrorMessage} from '@utils/errors';
+import {call, put} from 'redux-saga/effects';
+
+const actionCreator = AccountActions.GET_PERIOD_STATS.START.create;
+
```
ice-ares, line 10: rename everything to include user growth stats, cuz stats is not enough

src/store/modules/Stats/selectors/timeSeriesToGraphData.ts
```diff
@@ -0,0 +1,54 @@
...
+import {StatsPeriod, UsersBarGraphData} from '@store/modules/Stats/types';
+import {appLocale} from '@translations/i18n';
+import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc';
+
+dayjs.extend(utc);
+dayjs.locale(appLocale);
+
+export function getLabel(timeSeries: TimeSeries) {
+  return dayjs.utc(timeSeries.date).format('MM/DD');
```
ice-ares, line 13: u'd still to transform it into user's timezone right?

src/store/modules/Stats/selectors/timeSeriesToGraphData.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function getEmptyData(statsPeriod: StatsPeriod) {
+  return Array(statsPeriod)
+    .fill(null)
+    .map((_, i) => {
+      const date = new Date();
```
ice-ares, line 20: use only dayjs

src/screens/HomeFlow/UserGrowthGraph/index.tsx
```diff
@@ -10,19 +10,22 @@ import {useBottomTabBarOffsetStyle} from '@navigation/hooks/useBottomTabBarOffse
...
-  MOCK_TOTAL_USERS_GRAPH_DATA,
-} from '@screens/HomeFlow/Stats/components/UsersGrowthGraph/mockData';
+import {useGetBarGraphDataForStatsPeriod} from '@store/modules/Stats/hooks/useGetBarGraphDataForStatsPeriod';
+import {StatsPeriod} from '@store/modules/Stats/types';
 import {t} from '@translations/i18n';
 import React, {memo} from 'react';
 import {ScrollView, StyleSheet, View} from 'react-native';
 import {rem} from 'rn-units';
 
+const STATS_PERIOD: StatsPeriod = 30;
```
ice-ares, line 20: shouldn't this be from a dynamic list?

src/store/modules/Stats/selectors/timeSeriesToGraphData.ts
ice-hades: nope. here we transform it to utc timezone specifically

src/screens/HomeFlow/UserGrowthGraph/index.tsx
ice-hades: What do you mean? Like pass it as a screen parameter?
Now it is always 30, just moved it to a constant

src/store/modules/Stats/selectors/timeSeriesToGraphData.ts
ice-ares: isn't this what u display on the UI?

src/screens/HomeFlow/UserGrowthGraph/index.tsx
ice-hades: that is a different screen. we redirect to it only when 30 is selected from drop down. probably better pass it as a screen parameter here

src/store/modules/Stats/selectors/timeSeriesToGraphData.ts
ice-hades: @ice-ares @ice-orion do you agree? because otherwise I see we either get the collision in dates or a date from the future (when UTC is used)

src/store/modules/Stats/sagas/getPeriodStats.ts
```diff
@@ -0,0 +1,31 @@
...
+import {Api} from '@api/index';
+import {UserGrowth} from '@api/statistics/types';
+import {AccountActions} from '@store/modules/Stats/actions';
+import {getErrorMessage} from '@utils/errors';
+import {call, put} from 'redux-saga/effects';
+
+const actionCreator = AccountActions.GET_USER_GROWTH_STATS.START.create;
+
+export function* getPeriodStats(action: ReturnType<typeof actionCreator>) {
```
ice-ares, line 11: its user growth stats, not period stats. The period is the parameter.

src/screens/HomeFlow/UserGrowthGraph/index.tsx
ice-ares: the separate screen is for any of the selections in the dropdown, not just 30

src/store/modules/Stats/selectors/timeSeriesToGraphData.ts
ice-ares: all I care is that what the user sees is the date in their correct timezone. Everything else are implementation details. U can work internally only with UTC and display with timezone for the user

ice-ares (commented):
lets rename all those "statsPeriod" "stats" to userGrowthStats

src/store/modules/Stats/actions/index.ts
```diff
@@ -0,0 +1,18 @@
...
+const GET_USER_GROWTH_STATS = createAction('GET_USER_GROWTH_STATS', {
+  START: (statsPeriod: StatsPeriod) => ({statsPeriod}),
+  SUCCESS: (statsPeriod: StatsPeriod, userGrowth: UserGrowth) => ({
+    statsPeriod,
+    userGrowth,
+  }),
+  FAILED: (errorMessage: string) => ({errorMessage}),
+});
+
+export const AccountActions = Object.freeze({
```
ice-orion, line 16: Let's name it as `StatsActions`?

src/store/modules/Stats/reducer/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  total: 0,
+};
+
+function reducer(state = INITIAL_STATE, action: Actions): UserGrowthStatsState {
+  return produce(state, draft => {
+    switch (action.type) {
+      case AccountActions.GET_USER_GROWTH_STATS.SUCCESS.type: {
+        const {statsPeriod, userGrowth} = action.payload;
+        const newTimeSeriesStatsMap = {...state.timeSeriesStatsMap};
+        newTimeSeriesStatsMap[statsPeriod] = userGrowth.timeSeries;
```
ice-orion, line 31: We're using [immer](https://immerjs.github.io/immer/) for immutability, so you can simply mutate `draft`:
```
draft.timeSeriesStatsMap[statsPeriod] = userGrowth.timeSeries;
```

src/screens/HomeFlow/Stats/components/UsersGrowthGraph/index.tsx
```diff
@@ -27,20 +23,14 @@ export const SEGMENTS: ReadonlyArray<{text: string; key: GraphCategory}> = [
   {text: t('stats.active'), key: 'active'},
 ];
 
-const PERIODS = [
-  {label: t('periods.3_days'), length: 3},
-  {label: t('periods.7_days'), length: 7},
-  {label: t('periods.30_days'), length: 30, redirect: true},
-];
+const USER_GROWTH_STATS_PERIOD = 3;
```
ice-ares, line 26: DEFAULT_ USER_GROWTH_STATS_PERIOD

src/store/modules/Stats/types.ts
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {BarGraphData} from '@components/BarGraph/types';
+
+export type StatsPeriod = 3 | 7 | 14 | 30 | 90;
```
ice-ares, line 5: pick between this and this: src/store/modules/Stats/constants.ts 

We can't have the same thing defined in 2 places

src/store/modules/Stats/actions/index.ts
```diff
@@ -0,0 +1,18 @@
...
+const GET_USER_GROWTH_STATS = createAction('GET_USER_GROWTH_STATS', {
+  START: (statsPeriod: StatsPeriod) => ({statsPeriod}),
+  SUCCESS: (statsPeriod: StatsPeriod, userGrowth: UserGrowth) => ({
+    statsPeriod,
+    userGrowth,
+  }),
+  FAILED: (errorMessage: string) => ({errorMessage}),
+});
+
+export const UserGrowthStatesActions = Object.freeze({
```
ice-orion, line 16: Should it be `States` or `Stats`?

src/store/modules/Stats/types.ts
ice-ares: doesn't matter. If we want to add a new value, we shouldn't have to modify in 2 places

ice-orion: You can define array like so:
```
export const STATS_PERIODS = [3, 7, 14, 30, 90] as const;
```
and then pick the possible types for `StatsPeriod` like so:

```
export type StatsPeriod = typeof STATS_PERIODS[number];
```

FYI(optional): defining `UsersBarGraphData` type and assigning it seems to be redundant - if you remove it, TS will understand the types by itself and will keep the typesafety

ice-ares (dismissed):

src/components/BarGraph/index.tsx
```diff
@@ -1,43 +1,68 @@
...
 type Props = {
-  data: {label: string; value: number}[];
+  data: BarGraphData[];
 };
 
 const Y_AXIS_WIDTH = rem(36);
 const X_AXIS_HEIGHT = rem(20);
 const ROW_HEIGHT = rem(28);
 const NUMBER_OF_STEPS_X = 8;
+const FONT_SIZE = 10;
```
ice-hector, line 19: fontSize is usually less than lineHeight (that is 2-3 points bigger not to cut letters). 
I see you use it in calculations, should we use maybe lineHeight value instead?

src/screens/HomeFlow/Stats/components/Summary/index.tsx
```diff
@@ -32,7 +41,9 @@ export const Summary = memo(() => {
         />
         <View style={styles.cellDataRight}>
           <Text style={styles.labelText}>{t('stats.total')}</Text>
-          <Text style={styles.valueText}>139,205</Text>
+          <Text style={styles.valueText}>
+            {totalUsers.toLocaleString('en-US')}
```
ice-hector, line 45: Move 'en-US' to constants as DEFAULT_FORMAT_LOCALE and use this constant in all the places

src/components/BarGraph/index.tsx
```diff
@@ -1,43 +1,68 @@
...
-  const lastXValue = stepValue * NUMBER_OF_STEPS_X;
+  const stepValue = Math.floor(maxValue / NUMBER_OF_STEPS_X) || 1;
+  const lastXValue = Math.max(maxValue, stepValue * (NUMBER_OF_STEPS_X + 1));
+  const numberOfSteps =
+    Math.ceil(maxValue / stepValue) + 1 || NUMBER_OF_STEPS_X;
 
   const {transitionEnd} = useScreenTransitionEnd();
+  const [width, setWidth] = useState(100);
+
+  const footer = useMemo(() => {
```
ice-orion, line 35: It's better to use separate components and wrap it in `memo` when you need to create a memoized component, so here you can create a local `Footer` component.

* React.memo() is a higher-order component that we can use to wrap components that we do not want to re-render unless props within them change
* useMemo() is a React Hook that we can use to wrap functions within a component. We can use this to ensure that the values within that function are re-computed only when one of its dependencies change

They use different memoization mechanisms under the hood (memo returns special `REACT_MEMO_TYPE` component that is pocessed during the rendering instead of `useMemo` that simply returns previous value if the deps didn't change - and not guaranteed as they state)

src/hooks/useIsMounted.ts
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import * as React from 'react';
+
+export function useIsMounted() {
```
ice-orion, line 5: Build in prop `isMounted` was removed from react as antipattern and so we should also avoid the usage if it. It seems we can handle InteractionManger callback without this hook, so let's remove it

src/components/BarGraph/index.tsx
ice-hades: it's width is used here

src/components/BarGraph/index.tsx
```diff
@@ -1,43 +1,68 @@
...
 const ROW_HEIGHT = rem(28);
 const NUMBER_OF_STEPS_X = 8;
+const FONT_SIZE = 10;
 
 export const getBarGraphHeight = (numberOfRows: number) => {
   return X_AXIS_HEIGHT + numberOfRows * ROW_HEIGHT;
 };
 
-export const BarGraph = ({data}: Props) => {
+export const BarGraph = React.memo<Props>(({data}: Props) => {
```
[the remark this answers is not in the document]
ice-hades: Why using memo instead of React.memo would eliminate the need to pass Props?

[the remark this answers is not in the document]
ice-orion: The point is that you don't need to define it twice -> use either `memo` generic or define it explicitly for the function arguments:
```
memo<Props>(({data}) // pass in generic
memo(({data}: Props)  // set for func arguments
```

src/components/BarGraph/index.tsx
ice-hades: How can you prove one performs better than the other?
I saw lots of its usage that way. Why do you need to create a new component each time you want to memorise some render output?
Here is of the the usage examples - https://github.com/callstack/react-native-pager-view/blob/master/example/src/BasicPagerViewExample.tsx
it is even embedded here

src/constants.ts
```diff
@@ -0,0 +1,3 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+export const DEFAULT_FORMAT_LOCALE = 'en-US';
```
ice-hector, line 3: For constants we have a separate folder called src/constants/

src/components/BarGraph/index.tsx
ice-hades: @ice-orion performance is instant now for switching the states period - https://frameyapp.slack.com/archives/C0391DUR5GD/p1670492257411969?thread_ts=1670238181.107469&cid=C0391DUR5GD

ice-orion: I'm not telling that that is slow... I'm sure that in this particular case this would be instant even without this memo. By the way, on the video when you select 90, only a couple of rows are rendered, have you tried the performance when all the 90 rows have values?

ice-hades: Have you tried @ice-orion ? It was initially developed by you

this component should behave the same. matters only the number of rows not the data inside.
the question is how many concurrent animations would behave given animating the width is not recommended

ice-orion: No, i haven't tried 90 because when i developed it, the max number of rows was 30...

ice-orion (dismissed):

src/components/BarGraph/index.tsx
ice-hades: @ice-orion number of rows matters. I said data within those rows doesn't matter.

To be honest I would change some more things:

- I would definitely get rid of current animation which animates width. @ice-orion animating width leads to rerenders. only opacity and those within transform property can be changed directly in UI thread. So I would change it to `translateX` and hide bar initially on the left;
- I would replace Scroll Views with FlatLists. But that is not a trivial task in current codebase. Because there is a ScrollView on the top level we would need to flatten the render tree so we can replace that Scroll View on the top level with the Section View. All because it is not allowed (not recommended actually) to embed Flat Lists inside Scroll Views with the same orientation;

@ice-ares if we plan to display more then 90 days period stats then we probably need to check now how it would behave. And would be cool even now if we could populate 90 days back with some live data on staging so we can see how it would behave with real data not just a few recent days of data.

@ice-ares @ice-orion if time allows to spend us some more time on this I would definitely change at least the animations to be more performant

ice-ares: @ice-hades sure, spend more time on it

ice-hades: @ice-ares cool. so animations are must have. moving to flat list is only makes sense if we test it on real data and it shows slow performance. because that change would lead to flattening of the component tree. Could you populate the staging with real data? Now there are 0 active users

ice-orion: So it works with a big lag when all 90 rows have data

>  So I would change it to translateX and hide bar initially on the left;
Good idea!

>  would replace Scroll Views with FlatLists
This is always a good idea for big lists. And it's not that difficult to achieve on `UserGrowthGraph`: You can break `BarGraph` to separate components -> Footer and BarGraphRow. And then on the `UserGrowthGraph` screen you can have one FlatList with ListHeaderComponent=SectionHeader ListFooterComponent=BarGraphFooter and renderItem=BarGraphRow.
At the same time you can keep using `map` on the main Stats screen to render that 3 BarGraphRow.
This way you won't have FlatList inside ScrollView

src/App.tsx
```diff
@@ -4,12 +4,17 @@ import {Initialization} from '@components/Initialization';
...
-import {StatusBar, StyleSheet} from 'react-native';
+import {Platform, StatusBar, StyleSheet} from 'react-native';
 import {GestureHandlerRootView} from 'react-native-gesture-handler';
 import {SafeAreaProvider} from 'react-native-safe-area-context';
+import {enableFreeze} from 'react-native-screens';
 import {Provider} from 'react-redux';
 import {PersistGate} from 'redux-persist/integration/react';
 
+if (Platform.OS === 'android') {
+  enableFreeze();
```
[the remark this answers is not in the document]
ice-hades: It is explained here what it does - https://www.npmjs.com/package/react-native-screens
We use the same on Framey and we have some issue on ios so we disabled it for ios.
I checked the performance of user growth UI - seems like visually there are no gains from enabling it.
Probably no issue to remove it

ice-orion (approved):

ice-orion (dismissed):

src/App.tsx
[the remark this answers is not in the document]
ice-orion: You removed it in the prev commit and now got back. Is it by mistake or we do need it?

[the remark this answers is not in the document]
ice-hades: I think it is better to have it. If you have any objections let's remove it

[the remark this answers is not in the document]
ice-orion: No, i don't have any

ice-orion (approved):

ice-hector (approved):

PR #91, opened 2022-12-26
fix privacy modals nesting

ice-orion:
Fixed Privacy Modals nesting -> they should be in MainStack instead of ProfileStack to be on top of the bottom tabbar
Also fixed that weird white bottom line on android on the main profile screen


ice-orion:
Fixed Privacy Modals nesting -> they should be in MainStack instead of ProfileStack to be on top of the bottom tabbar
Also fixed that weird white bottom line on android on the main profile screen

ice-ares (approved):

ice-hades (approved):

PR #97, opened 2023-01-05
Feature/login flow fixes

ice-hector:


.github/workflows/CICD.yaml
```diff
@@ -69,7 +69,6 @@ jobs:
         run: |
           cd main-repo
           yarn
-          yarn fetch-translations
```
ice-orion, line 72: rebase with master please

src/screens/WelcomeFlow/ClaimUsername/hooks/useClaimUsername.ts
```diff
@@ -37,53 +41,92 @@ export const useClaimUsername = () => {
...
+  const resetError = useCallback(() => {
+    if (validationError) {
+      dispatch(ValidationActions.USERNAME_VALIDATION.RESET.create());
     }
-  };
+  }, [validationError, dispatch]);
+
+  const updateUsername = useCallback(
+    (newUsername: string) => {
+      if (newUsername.length < 4) {
```
ice-ares, line 67: Make those into constants

ice-ares (commented):
Cant see where u are handling errors from firebase auth for that invalid link UI

src/hooks/useAsyncStorage.ts
```diff
@@ -0,0 +1,31 @@
...
+    } catch (error) {
+      logError(error);
+    }
+  };
+
+  const removeStored = () => {
+    AsyncStorage.removeItem(key);
+  };
+
+  const setStored = async (value: any) => {
```
ice-ares, line 22: use this to delete item as well, if value is null

src/screens/WelcomeFlow/Onboarding/hooks/useFinishOnboarding.ts
```diff
@@ -1,44 +1,76 @@
...
-            },
-          },
-          function* (freshUser) {
-            finishOnboarding(freshUser);
-            return {retry: false};
-          },
-        ),
-      );
+  const finishOnboarding = async () => {
+    if (!onboardingIds.includes(deviceUniqueId)) {
```
ice-ares, line 43: huh? this is a device storage, why are u doing that? there is only 1 device

src/translations/translations.json
```diff
@@ -1,732 +1,752 @@
 {
-    "ro": {
-        "settings": {
```
ice-ares, line 3: why did u do this?

src/screens/WelcomeFlow/Onboarding/hooks/useFinishOnboarding.ts
ice-hector: Yeah, my bad, I had the idea to store device ids if we save on the backend side but we switched to async storage and local storing. Here should be userId.

src/translations/translations.json
ice-hector: Do what? Diff just displayed in a strange way like a lot of lines were deleted and then added. From translation file should be correct

src/navigation/options.ts
```diff
@@ -9,6 +9,7 @@ export const tabOptions = {
 
 export const screenOptions = {
   headerShown: false,
+  gestureEnabled: false,
```
ice-orion, line 12: why?

src/screens/AuthFlow/ConfirmEmailLink/hooks/useConfirmEmailLink.ts
```diff
@@ -43,6 +47,12 @@ export const useConfirmEmailLink = () => {
     }, [dispatch]),
   );
 
+  useEffect(() => {
+    if (validateError) {
+      navigation.replace('InvalidLink');
+    }
+  }, [validateError, navigation]);
```
ice-orion, line 54: Shouldn't we rely on error code? Or `InvalidLink` screen is suitable for any possible validation error?

src/screens/AuthFlow/InvalidLink/components/Header/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import React from 'react';
+import {Image, StyleSheet, Text, View} from 'react-native';
+import {rem, screenWidth} from 'rn-units';
+
+export const Header = () => {
+  const topOffset = useTopOffsetStyle();
+  return (
+    <View style={topOffset.current}>
+      {/* for too tall devices */}
+      <View style={styles.overscrollTop} />
```
ice-orion, line 18: I think you don't need this since you don't have scroll on this screen

src/screens/AuthFlow/InvalidLink/hooks/useInvalidLink.ts
```diff
@@ -0,0 +1,52 @@
...
+        },
+      );
+      return () => subscription.remove();
+    }, [dispatch]),
+  );
+
+  return {
+    email,
+    validateError,
+    validateLoading,
```
ice-orion, line 49: It seems you don't use `email` and `validateError` from this hook, let's remove them.
+ why do we have `validateLoading` and loading animation on the InvalidLink screen, what's the case when this loading is `true`?

src/screens/AuthFlow/SignIn/index.tsx
```diff
@@ -21,6 +19,7 @@ import {Tab, Tabs} from '@screens/AuthFlow/SignIn/components/Tabs';
 import {useEmailAuth} from '@screens/AuthFlow/SignIn/hooks/useEmailAuth';
 import {usePhoneAuth} from '@screens/AuthFlow/SignIn/hooks/usePhoneAuth';
 import {useSocialAuth} from '@screens/AuthFlow/SignIn/hooks/useSocialAuth';
+import {DIVIDER_HEIGHT} from '@screens/HomeFlow/Home/components/Overview/components/AdoptionCard/components/LevelRow';
```
ice-orion, line 22: I think you import it from a wrong place

src/screens/WelcomeFlow/ClaimUsername/hooks/useClaimUsername.ts
```diff
@@ -37,53 +42,92 @@ export const useClaimUsername = () => {
...
+  const resetError = useCallback(() => {
+    if (validationError) {
+      dispatch(ValidationActions.USERNAME_VALIDATION.RESET.create());
     }
-  };
+  }, [validationError, dispatch]);
+
+  const updateUsername = useCallback(
+    (newUsername: string) => {
+      if (newUsername.length < USERNAME_MIN_LENGTH) {
```
ice-orion, line 68: Let's have all the validations in sagas so we won't need to duplicate it (e.g. we also may edit username in settings so we also need validations there. If we have that logic in saga, it would be handled at one place).

And same for the `whoInvitedYou` hook

src/screens/WelcomeFlow/ClaimUsername/hooks/useClaimUsername.ts
```diff
@@ -37,53 +42,92 @@ export const useClaimUsername = () => {
...
+          AccountActions.UPDATE_ACCOUNT.START.create(
+            {
+              username: username,
+              clientData: {
+                ...user.clientData,
+                registrationProcessFinalizedSteps: steps,
+              },
+            },
+            function* (freshUser) {
+              if (freshUser.username !== username) {
```
ice-orion, line 120: `username` might be that same, but you also modify the `finalizedSteps`, so the conflict might be there. Also you're calling `updateUsername` in this case that first calls validate, then wait's for a second and only then calls update once again. You can just call update right away, since you already passed the validation with this username. Take a look how it's made in other places (separate function that accepts user and modified fields and then we pass freshUser in recursion in this callback)

src/screens/WelcomeFlow/ConfirmEmailCode/index.tsx
```diff
@@ -45,6 +45,7 @@ export const ConfirmEmailCode = () => {
           errorText={validationError}
           editable={!validateLoading}
           validated={isSuccessValidation}
+          selectionColor={'green'}
```
ice-orion, line 48: COLORS.shamrock?

src/screens/WelcomeFlow/IceBonus/index.tsx
```diff
@@ -19,8 +21,25 @@ export const IceBonus = () => {
       header={
         <BigHeader
           title={t('ice_bonus.title')}
-          description={t('ice_bonus.description', {value: currentBalance})}
-          progressPercentage={100}
+          description={
+            <>
+              {t('ice_bonus.description_part1', {
+                value: Number(currentBalance),
```
ice-orion, line 27: Why Number? I think it will be string that we will receive from the BE.

+ let's use `reactStringReplace` and one translation tag instead of breaking it on separate parts. (take a look at `staking.terms_agree` translation and `reactStringReplace` usage for it)

src/screens/WelcomeFlow/Onboarding/hooks/useFinishOnboarding.ts
```diff
@@ -1,44 +1,75 @@
...
+      await setStored(updatedDeviceIds);
     }
   };
 
-  const onSubmit = () => finishOnboarding(user);
+  const onSubmit = () => finishOnboarding();
 
-  return {onSubmit, loading};
+  const canAskNotificationPermissions = async () => {
+    const result = await checkNotifications();
```
ice-orion, line 51: Why do we have it here? We have permissions module for that and you can use `permissionSelector` to grab the status on the push permissions.

src/screens/WelcomeFlow/Onboarding/hooks/useFinishOnboarding.ts
```diff
@@ -1,44 +1,75 @@
...
 
 export const useFinishOnboarding = () => {
-  const dispatch = useDispatch();
+  const [slides, setSlides] = useState<OnboardingSlide[]>([]);
   const user = useSelector(userSelector) as User;
+  const {setStored, getStored} = useAsyncStorage('onboarding', []);
+  const [onboardingIds, setOnboardingIds] = useState<string[]>([]);
+
+  useEffect(() => {
+    canAskNotificationPermissions().then(result => {
```
ice-orion, line 23: use permissions module

src/screens/WelcomeFlow/Onboarding/slides.tsx
```diff
@@ -2,36 +2,38 @@
...
 import {IceLabel} from '@components/Labels/IceLabel';
 import {t} from '@translations/i18n';
-import React from 'react';
-import {StyleSheet, Text} from 'react-native';
+import React, {ReactNode} from 'react';
+import {ImageSourcePropType, StyleSheet, Text} from 'react-native';
 import {isAndroid} from 'rn-units';
 
+const ICON_Y_OFFSET = isAndroid ? 6 : 2;
+const ICON_SIZE = 23;
```
ice-orion, line 10: rem

src/screens/WelcomeFlow/SetEmail/hooks/useSetEmail.ts
```diff
@@ -26,6 +29,25 @@ export const useSetEmail = () => {
...
+    removeRefStep();
+  };
+
+  const removeRefStep = () => {
+    let finalizedSteps =
+      user.clientData?.registrationProcessFinalizedSteps ?? [];
+    if (finalizedSteps.includes('referral')) {
+      finalizedSteps = finalizedSteps.filter(step => step !== 'referral');
+      dispatch(
+        AccountActions.UPDATE_ACCOUNT.START.create({
```
ice-orion, line 42: add race condition callback

src/screens/WelcomeFlow/WhoInvitedYou/hooks/useWhoInvitedYou.ts
```diff
@@ -43,59 +44,101 @@ export const useWhoInvitedYou = () => {
...
-          {
-            ...(referredBy ? {referredBy} : {}),
-            clientData: {
-              ...currentUser.clientData,
-              registrationProcessFinalizedSteps: [
-                ...finalizedSteps,
-                'referral',
-              ],
-            },
+        AccountActions.UPDATE_ACCOUNT.START.create({
```
ice-orion, line 72: race condition callback

src/utils/string.ts
```diff
@@ -21,3 +21,5 @@ export const stringToColor = (input: string = '') => {
 export const removeInvalidUsernameCharacters = (input: string) => {
   return input.replace(/[^a-zA-Z0-9-]/g, '');
 };
+
+export const usernameValidationRegex: RegExp = /^[a-zA-Z0-9-]+$/;
```
ice-orion, line 25: may be it would be better to have `validateUsername` function that would use this regexp to check the username argument since this is the only possible use case for this regexp anyway?

src/screens/WelcomeFlow/Onboarding/hooks/useFinishOnboarding.ts
```diff
@@ -1,44 +1,99 @@
...
 
-  const finishOnboarding = (currentUser: User) => {
-    const finalizedSteps =
-      currentUser?.clientData?.registrationProcessFinalizedSteps ?? [];
-    if (!finalizedSteps.includes('onboarding')) {
-      dispatch(
-        AccountActions.UPDATE_ACCOUNT.START.create(
-          {
+  const finishOnboarding = useCallback(() => {
+    console.log('finishOnboarding');
```
ice-ares, line 32: remove all console logs

src/translations/translations.json
```diff
@@ -1,732 +1,753 @@
 {
-    "ro": {
```
ice-ares, line 2: you need to make sure u’re not indenting differently, cuz i cant see what changed

src/navigation/options.ts
ice-hector: Because we don't allow to go back there but you can do this with swipe gesture if its set to true

src/screens/AuthFlow/ConfirmEmailLink/hooks/useConfirmEmailLink.ts
ice-hector: Yeah, for any email link validation error we show this screen, that was the requirement

src/utils/string.ts
ice-hector: Done

src/translations/translations.json
ice-hector: I'm using auto-formatting, so there are not many options. From the key/values side, all should be good because we covered localization with types so if the project uses some localization keys that do not exist we will definitely see this in the commit error

src/hooks/useAsyncStorage.ts
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import AsyncStorage from '@react-native-async-storage/async-storage';
+import {logError} from '@services/logging';
+
+export const useAsyncStorage = (key: string, initialValue: any) => {
```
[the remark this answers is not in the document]
ice-hector: I fully rewrote that storage and removed the default value from there, and added types

src/navigation/Welcome.tsx
```diff
@@ -37,39 +38,74 @@ export type WelcomeStackParamList = {
...
+  'IceBonus',
+];
+
 export function WelcomeNavigator() {
+  const initializedRef = useRef(false);
   const user: ReturnType<typeof userSelector> = useSelector(userSelector);
   const emailVerificationStep = useSelector(emailVerificationStepSelector);
   const navigation =
     useNavigation<NativeStackNavigationProp<WelcomeStackParamList>>();
+  const {getStored} = useAsyncStorage('onboarding', []);
```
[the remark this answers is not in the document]
ice-hector: Moved to constants/persist

src/screens/AuthFlow/InvalidLink/hooks/useInvalidLink.ts
ice-hector: There is no case for validateLoading. Also removed same as email and validateError

src/screens/WelcomeFlow/IceBonus/index.tsx
ice-hector: Because on the backend side it's a string with .00 (Example: 17.00). We need only an integer according to design.
Translation string supports both (number | string) so in this case Number() just converts the value to an integer in our case.

As for reactStringReplace, are you sure that we should use it? I checked `staking.terms_agree` and It's VERY hard to read it (same as ternary expressions, looks like not readable overhead to me).

src/screens/WelcomeFlow/ClaimUsername/hooks/useClaimUsername.ts
ice-hector: Moved to sagas

src/hooks/useAsyncStorage.ts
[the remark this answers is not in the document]
ice-hector: I remained it as a hook but refactored it, can you check pls? 
I don't want to rewrite it right now just because we can "potentially" use it in sagas sometime in the future, there were no such requirements

src/navigation/Welcome.tsx
```diff
@@ -37,39 +38,74 @@ export type WelcomeStackParamList = {
...
+  const [showOnboarding, setShowOnboarding] = useState(true);
+
+  useEffect(() => {
+    getStored().then(onboardingIds => {
```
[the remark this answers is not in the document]
ice-hector: You couldn't return null instead of navigator... Any other options?

src/hooks/useAsyncStorage.ts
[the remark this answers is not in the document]
ice-orion: My point here is that you don't need to wrap it in `memo`, cuz those 2 functions have nothing to do with react and you can just export those as is:
```
export const getStored = async (): Promise<T | undefined> => {
        try {
          const item = await AsyncStorage.getItem(key);
          return item ? JSON.parse(item) : undefined;
        } catch (error) {
          logError(error);
        }
      }
      
export const setStored = async (value: T) => {
        try {
          if (value) {
            await AsyncStorage.setItem(key, JSON.stringify(value));
          } else {
            await AsyncStorage.removeItem(key);
          }
        } catch (error) {
          logError(error);
        }
      }
      ```
      
And that's it - no need to rewrite anything, just get rid of unnecessary complications here.

File: src/screens/WelcomeFlow/IceBonus/index.tsx (hunk 14 above)
Comment (ice-orion) [reply src/screens/WelcomeFlow/IceBonus/index.tsx:27]:
```
Number('1,234.00') -> NaN
```
I would clarify if we really have to cut off the decimals, cuz everywhere else we show it as is.

File: src/components/AnimatedSplash/index.tsx [hunk 26]
```diff
@@ -0,0 +1,60 @@
...
+import {isAppInitializedSelector} from '@store/modules/AppCommon/selectors';
+import AnimatedLottieView from 'lottie-react-native';
+import React, {useState} from 'react';
+import {useEffect} from 'react';
+import {StyleSheet, View} from 'react-native';
+import {useSelector} from 'react-redux';
+import {screenWidth} from 'rn-units';
+
+export const AnimatedSplash = () => {
+  const isAppInitialized = useSelector(isAppInitializedSelector);
```
ice-orion, line 14: We have this code and component in the Router:
```
if (!isAppInitialized) {
    return <Initialization />;
  }
```

It seems it can be:
```
if (!isAppInitialized) {
   return null
}
```
And the `Initialization` component can be removed?

src/navigation/Welcome.tsx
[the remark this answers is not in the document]
ice-orion: Well, it seems we should have used redux persist here, then reading `onboardingIds` wouldn't be async here and everything would have been fine... But ok, if there are no straight ways to handle it and the screen change is not noticeable then close the thread

src/screens/WelcomeFlow/SetEmail/hooks/useSetEmail.ts
```diff
@@ -26,6 +29,31 @@ export const useSetEmail = () => {
     setEmail(text);
   };
 
+  const onBack = () => {
+    removeRefStep(user);
+  };
+
+  const removeRefStep = (userToUpdate: User) => {
```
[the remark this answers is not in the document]
ice-hector: Otherwise, front navigation will not work if we go back. All these finalized steps storing on backend to restore something is tricky implementation. On my mind its much better to make a decision depending on user model fields (like you have not email -> you show Set email screen, you didnt get ice bonus -> you show ice bonus screen. Same for username / reffered by/ etc.)

src/components/AnimatedSplash/index.tsx
ice-hector: Yeah, good catch

src/screens/WelcomeFlow/ClaimUsername/hooks/useClaimUsername.ts
```diff
@@ -36,54 +34,91 @@ export const useClaimUsername = () => {
...
-      debounce((text: string) => {
-        if (text) {
-          dispatch(ValidationActions.USERNAME_VALIDATION.START.create(text));
-        }
-      }, 600),
-    [dispatch],
-  );
+  const onSubmit = () => {
+    Keyboard.dismiss();
+    if (username !== user.username) {
```
ice-ares, line 42: Equal ignore csse

src/screens/WelcomeFlow/WhoInvitedYou/hooks/useWhoInvitedYou.ts
```diff
@@ -34,69 +36,116 @@ export const useWhoInvitedYou = () => {
...
+
   const updateError = useSelector(
     failedReasonSelector.bind(null, AccountActions.UPDATE_ACCOUNT),
   );
   const updateLoading = useSelector(
     isLoadingSelector.bind(null, AccountActions.UPDATE_ACCOUNT),
   );
 
-  const [refUsername, setRefUsername] = useState('');
+  const initialRefUsername = useRef(user.referredBy ?? null);
```
ice-ares, line 53: ReferredBy is not username

src/screens/WelcomeFlow/WhoInvitedYou/hooks/useWhoInvitedYou.ts
```diff
@@ -34,69 +36,116 @@ export const useWhoInvitedYou = () => {
...
+      }
+    },
+    [dispatch],
+  );
+
+  const onSubmit = () => {
+    Keyboard.dismiss();
+    resetError();
+    dispatch(
+      AccountActions.UPDATE_ACCOUNT.START.create({referredBy: refUsername}),
```
ice-ares, line 147: ReferredBy is userId

src/services/auth/index.ts
```diff
@@ -133,6 +133,10 @@ export const getAuthenticatedUser = async (forceRefresh?: boolean) => {
 export const getAuthErrorMessage = (error: unknown) => {
   if (checkProp(error, 'code')) {
     switch (error.code) {
+      case 'auth/code-expired':
+        return t('errors.code_expired');
```
ice-ares, line 137: You dont display that, you just retry

src/services/auth/index.ts
```diff
@@ -133,6 +133,10 @@ export const getAuthenticatedUser = async (forceRefresh?: boolean) => {
...
+      case 'auth/too-many-requests':
+        return t('errors.too_many_requests');
```
ice-ares, line 139: U dont display that, u just retry

src/store/modules/Account/sagas/updateAccount.ts
```diff
@@ -33,6 +35,18 @@ export function* updateAccountSaga(action: ReturnType<typeof actionCreator>) {
       throw {code: Errors.InvalidEmail};
     }
 
+    if ('referredBy' in userInfo) {
+      const refUser: SagaReturnType<typeof validateRefUsername> = yield call(
```
ice-ares, line 39: No, u already have it when u validate the input, take it from there, dont call it again

src/store/modules/Account/sagas/updateAccount.ts
```diff
@@ -100,3 +124,43 @@ export function* updateAccountSaga(action: ReturnType<typeof actionCreator>) {
     throw error;
   }
 }
+
+function* validateUsername(username: string) {
```
ice-ares, line 128: No, u dont need those, u have everything u need already

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
```diff
@@ -1,47 +1,39 @@
...
 
 export function* validateRefUsernameSaga(
   action: ReturnType<typeof actionCreator>,
 ) {
   const {username} = action.payload;
-  const user: ReturnType<typeof userSelector> = yield select(userSelector);
 
-  try {
-    const refUser: SagaReturnType<typeof Api.user.getUserByUsername> =
-      yield call(Api.user.getUserByUsername, {username});
```
ice-ares, line 21: Bad idea to move this out of here and into updateAccount

src/store/modules/Validation/sagas/validateUsernameSaga.ts
```diff
@@ -1,42 +1,37 @@
...
 
 const actionCreator = ValidationActions.USERNAME_VALIDATION.START.create;
 
 export function* validateUsernameSaga(
   action: ReturnType<typeof actionCreator>,
 ) {
   const {username} = action.payload;
-  try {
-    const validationResult: SagaReturnType<typeof Api.user.getUserByUsername> =
-      yield call(Api.user.getUserByUsername, {username});
```
ice-ares, line 18: Same thing here

src/translations/locales/en.json
```diff
@@ -1,32 +1,31 @@
...
     "page2": {
       "title": "Planet Friendly",
-      "description": " is inherently more energy efficient than BTC\nas it uses a “Proof of Stake” consensus mechanism.\nWith proof-of-stake (POS), cryptocurrency\nowners validate block transactions based on the\nnumber of coins a validator stakes."
+      "description": " is inherently more energy efficient than BTC as it uses a “Proof of Stake” consensus mechanism. With proof-of-stake (POS), cryptocurrency owners validate block transactions based on the number of coins a validator stakes."
     },
     "page3": {
       "title": "People’s Coin",
-      "description_part1": "The power is in your hands. Everything that we develop at ",
-      "description_part2": " will be open source & controlled by the Decentralized Autonomous Organization (DAO). "
+      "description_part1": "The power is in your hands.\nEverything that we develop at ",
```
ice-ares, line 13: Dont do that

src/translations/locales/en.json
```diff
@@ -1,32 +1,31 @@
...
       "title": "Stay Connected",
       "description_part1": "Tap ",
-      "description_part2": " and start mining today with our free, \nenergy-light mobile app. Your daily tap ensures\nthat ",
-      "description_part3": " ends up in the hands of real\nhumans, not bots."
+      "description_part2": " and start mining today with our free, energy-light mobile app. Your daily tap ensures that ice ends up in the hands of real humans, not bots."
     },
     "page5": {
       "title": "Refer & Earn",
-      "description": "Invite your friends and earn(both of you) up to\n25% bonus on your base mining rate for\neach friend that joins your team."
+      "description": "Invite your friends and earn (both of you) up to 25% bonus on your base mining rate for each friend that joins your team."
```
ice-ares, line 23: Every number is a parameter, not hardcoded

ice-hector: Should I hardcode it or is this one dynamic and can be gotten from the backend?

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
ice-hector: Why it's a bad idea? The flow has been changed. Now we have two layers of validation
1) Local validations 
2) Backend-related validations

So yeah, getting ref user now located in Update_account (same as Claim username backend-related part)
There are several advantages of using such an approach and we discussed this on a team call.
1) We have different places in the app where we do pretty same things (claim username in auth flow vs change username in settings and so on). From now we simply call first the local validations reactively when the user types and then just the update profile (on some user action) which already contains the backend-related validations.
2) We change the user field (like username). Why is it bad to validate it in the update account saga? It's all fully related.

src/store/modules/Validation/sagas/validateUsernameSaga.ts
ice-hector: Same answer as here
https://github.com/ice-blockchain/mobile-app/pull/97#discussion_r1089934015

src/store/modules/Account/sagas/updateAccount.ts
ice-hector: Looks like you didn't understand how the new validation flow works. We separated the local one and backend-related one.

src/screens/WelcomeFlow/ClaimUsername/hooks/useClaimUsername.ts
ice-hector: So what is the issue here? If they equal I just skip validation and do` updateFinalizedSteps(user);` in else

src/store/modules/Account/sagas/updateAccount.ts
ice-hector: No, I don't have, flow has been changed. 
https://github.com/ice-blockchain/mobile-app/pull/97#discussion_r1089934015

src/translations/locales/en.json
ice-ares: Const everything, but get them out of the translations

ice-hector: Yeah ok, I refactoring this translations part right now to support our new approach

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
ice-ares: Cuz you messed up the separation of concerns. Validation saga is the obly one that validates, update saga is the only pne that updates. 

Moving all the validation login to those validation sagas fixes that

ice-ares: @ice-orion fyi

src/store/modules/Account/sagas/updateAccount.ts
ice-ares: I know, but the code doesn’t have to be like that, u can run all ur FE validation logic, in the validate saga, before calling the BE

src/screens/WelcomeFlow/ClaimUsername/hooks/useClaimUsername.ts
ice-ares: Ignore case when u compare em

src/store/modules/Account/sagas/updateAccount.ts
ice-hector: I do exactly this way. The only difference is that in the validation saga, we do all local validations and we do them reactively when the user is typing. All other validations/errors (that basically, we know about from BE requests are moved to updateAccountSaga, because we already handle there tens other errors, so there is no reason not to handle several others there

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
ice-hector: I disagree. Please check this comment
https://github.com/ice-blockchain/mobile-app/pull/97#discussion_r1089970832

src/screens/WelcomeFlow/IceBonus/index.tsx
ice-hector: Yeah we need to cut, that's the design requirement, we have there 00  near in fully different size and style

src/store/modules/Account/sagas/updateAccount.ts
ice-ares: The GetUserByUsername u use to validate referral returns an user, with an id, use that id as the referredBy

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
ice-orion: Generally for me it also looks fine having validations as part of the update logic. If we have those separated, we can potentially bypass validations and call update with a corrupted data. On the other hand if we have the same checks during the update on the BE side, then probably we can remove it from here. 

So we can probably remove `validateUsername` function where we check that username is already taken since the update function does the same (+ we even already handle the error).

So only `validateRefUsername` left and here the problem exists with the way we pass `referredBy`. So from the UI we're setting `referredBy` as username (since this is what user types). Then in saga we're taking username, fetching a user, taking user's id and setting back this id as `referredBy`. This looks confusing. Here is how we can fix it:
1. Pass the ref's username as a separate field. E.g. `referredByUsername`
2. Add `referredByUsername` field to the user's entity on the BE side, so we can set it during the update and BE do the verifications and populate `referredBy` with the ref's id
3. (probably the most clean way) We can have a separate action and saga for updating ref. If would accept `refUsername`, validate that the user exists, pick id and call `updateUser` with the correct `referredBy` field.

ice-ares: I am ok with what @ice-orion suggests

ice-ares: @ice-orion resolve this if/when u see fit

src/screens/WelcomeFlow/IceBonus/index.tsx
ice-orion: use <FormattedNumber />

ice-hector: @ice-orion please fix in scope of you backend rest api connection task

src/store/modules/Validation/sagas/validateUsernameSaga.ts
```diff
@@ -1,42 +1,37 @@
...
 import {ValidationActions} from '@store/modules/Validation/actions';
 import {t} from '@translations/i18n';
-import {getErrorMessage, showError} from '@utils/errors';
-import {call, put, SagaReturnType} from 'redux-saga/effects';
+import {validateUsername} from '@utils/string';
+import {put} from 'redux-saga/effects';
 
 const actionCreator = ValidationActions.USERNAME_VALIDATION.START.create;
 
 export function* validateUsernameSaga(
```
ice-atlas, line 11: Why do we need to keep it as saga, if it can be just plain function for validation? (if we remove request to BE)

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
```diff
@@ -1,47 +1,47 @@
...
 import {ValidationActions} from '@store/modules/Validation/actions';
 import {t} from '@translations/i18n';
-import {getErrorMessage, showError} from '@utils/errors';
-import {call, put, SagaReturnType, select} from 'redux-saga/effects';
+import {validateUsername} from '@utils/string';
+import {put, select} from 'redux-saga/effects';
 
 const actionCreator = ValidationActions.REF_USERNAME_VALIDATION.START.create;
 
 export function* validateRefUsernameSaga(
```
ice-atlas, line 12: Same here `Why do we need to keep it as saga, if it can be just plain function for validation? (if we remove request to BE)`

Any significant difference from `validateUsernameSaga `?

src/store/modules/Permissions/selectors/index.ts
```diff
@@ -4,7 +4,12 @@ import {PermissionType} from '@store/modules/Permissions/sagas/getPermissionsSag
 import {RootState} from '@store/rootReducer';
 import {RESULTS} from 'react-native-permissions';
 
-export const permissionSelector =
+export const grantedPermissionSelector =
```
ice-atlas, line 7: Maybe better to use `is...` naming for `bool` selectors? Example `isPermissionGrantedSelector`

src/store/configureStore.ts
```diff
@@ -26,9 +45,12 @@ export const store = configureStore({
         ignoredPaths: [
           'utilityProcessStatuses.UPDATE_ACCOUNT.payload.raceConditionStrategy',
         ],
+        ignoredActions: [FLUSH, REHYDRATE, PAUSE, PERSIST, PURGE, REGISTER],
```
ice-atlas, line 48: Why we ignore all this actions?

src/store/configureStore.ts
```diff
@@ -1,12 +1,29 @@
...
+} from 'redux-persist';
 import createSagaMiddleware from 'redux-saga';
 
 import {rootReducer} from './rootReducer';
 import {rootSaga} from './rootSaga';
 
+const persistConfig = {
+  key: 'root',
+  storage: AsyncStorage,
+  whitelist: ['users'],
```
ice-atlas, line 24: Also you can implement persisting on `usersReducer` level (whitelist only needed fields)

src/store/modules/Account/sagas/updateRefByUsernameSaga.ts
```diff
@@ -0,0 +1,44 @@
...
+  const {refUsername} = action.payload;
+  try {
+    const userId: ReturnType<typeof userIdSelector> = yield select(
+      userIdSelector,
+    );
+    const refUser: SagaReturnType<typeof Api.user.getUserByUsername> =
+      yield call(Api.user.getUserByUsername, {username: refUsername});
+    if (refUser.id === userId) {
+      yield put(
+        AccountActions.UPDATE_REF_BY_USERNAME.FAILED.create(
```
ice-atlas, line 25: just `throw` error here with errorMessage. SO you do not need `else` just `SUCCESS` in the end of `OK` flow

ice-atlas (changes requested):

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
ice-hector: Yeah, big difference, in validateUsernameSaga there are only local validation, this one (validateRefUsernameSaga) is like a middleware between local validations and Update_account call.

src/store/configureStore.ts
ice-hector: It part of rehydration integration was  in the required steps, or their official recommendation to fix some issue, don't remember

src/store/configureStore.ts
ice-hector: As I understood  from their dox redux persist doesn't support deep tree white listing

src/store/modules/Permissions/selectors/index.ts
ice-hector: We try to avoid is... naming

src/store/modules/Validation/sagas/validateUsernameSaga.ts
ice-hector: It couldn't be the plain function, because we listen to action states/results, reset errors when needed (that affects UI), and so on.  So to keep things consistent that's also done using the same approach

src/translations/locales/en.json
```diff
@@ -517,7 +514,7 @@
     "no_notifications": "No notifications to display",
     "add_announcements": "Add mocked announcements",
     "add_both": "Add mocked notifications \nand announcements together",
-    "clear_title": "Clear all notificaitons? ",
+    "clear_title": "Clear all notifications? ",
```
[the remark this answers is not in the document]
ice-hector: Dont know where it came from, not the code in scope of my tasks, but I thing we gonna definitely recheck all the localization and remove all the rest places with \n and things like this in scope of one separate task

src/App.tsx
```diff
@@ -17,10 +19,13 @@ export function App() {
...
       <Provider store={store}>
-        <StatusBar translucent backgroundColor="transparent" />
-        <GestureHandlerRootView style={styles.container}>
-          <Router />
-        </GestureHandlerRootView>
+        <PersistGate loading={null} persistor={persistor}>
+          <StatusBar translucent backgroundColor="transparent" />
+          <GestureHandlerRootView style={styles.container}>
+            <Router />
+            <AnimatedSplash />
```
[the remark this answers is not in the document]
ice-hector: I have to check and retest how it behave on both platforms if I move it outside

src/translations/locales/en.json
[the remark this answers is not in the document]
ice-hector: Fixed

src/store/modules/Account/sagas/updateRefByUsernameSaga.ts
ice-hector: I removed if statement at all, it's not possible to get it because we moved the refer itself validation when the user is typing, so next button will just be disabled for that case

src/screens/WelcomeFlow/WhoInvitedYou/hooks/useWhoInvitedYou.ts
```diff
@@ -13,90 +13,141 @@ import {
...
   const updateError = useSelector(
     failedReasonSelector.bind(null, AccountActions.UPDATE_ACCOUNT),
   );
   const updateLoading = useSelector(
     isLoadingSelector.bind(null, AccountActions.UPDATE_ACCOUNT),
   );
 
+  const initialReferredBy = useRef(user.referredBy);
   const [refUsername, setRefUsername] = useState('');
+  const [updateAccountFinished, setUpdateAccountFinished] = useState(false);
```
ice-orion, line 51: just out of curiosity - in `useClaimUsername` you rely on `isSuccessUpdate` to show the green checkmark on the text input, but here you store this flag in state. Can't we handle it the same way in both places? Or there are some differences?

src/store/configureStore.ts
ice-orion: `+ 1`
https://github.com/rt2zz/redux-persist#nested-persists
persist only 1 flag by wrapping 
```
export const usersReducer = persistReducer({whitelist: ....}, reducer),;
```

src/store/modules/Permissions/selectors/index.ts
ice-orion: In this particular case i would also say that `isPermissionGrantedSelector` is more clear. As for the naming ... It's not a rule to avoid `is` in the bool variable naming, but it's not a something required as well

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
ice-atlas: but the rule for naming almost the same, maybe we can reuse it some how? It may lead to bug if we need to change validation rule in one place we may forget about another

src/store/modules/Validation/sagas/validateUsernameSaga.ts
ice-atlas: any problem to have it as `hook`?

ice-hector: We already have related hooks/validations, that is the level above all that

src/store/modules/Validation/sagas/validateRefUsernameSaga.ts
ice-hector: My bad didn't read the first question properly, That is two different things related to different scenarios (username and refUsername are not the same). So its nothing to change here in that sense

src/store/configureStore.ts
ice-hector: Fixed

src/screens/AuthFlow/SignIn/index.tsx
```diff
@@ -86,15 +80,15 @@ export const SignIn = () => {
...
-              activeTab === 'phone' ? smsSentTimestamp : emailSentTimestamp
-            }
           />
           <Divider />
           <SocialButtons />
         </View>
+        <PrivacyTerms />
       </ScrollView>
-      {isSocialAuthLoading && <FullScreenLoading />}
+      {isSocialAuthLoading && selectedSocialType !== 'apple' && (
```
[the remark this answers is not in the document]
ice-hector: Yep, for apple we don't show loading, it has its own dark overlay

src/screens/WelcomeFlow/Onboarding/components/OnboardingSlide/index.tsx
```diff
@@ -31,6 +31,7 @@ const styles = StyleSheet.create({
   description: {
     flex: 1,
     marginHorizontal: rem(24),
+    zIndex: 1,
```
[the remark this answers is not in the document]
ice-hector: Because sometimes on low res screens text goes under the bottom image. This way its always on top of it

src/components/PrivacyTerms/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+import {LINKS} from '@constants/links';
+import {t} from '@translations/i18n';
+import {openLinkWithInAppBrowser} from '@utils/device';
+import {font} from '@utils/styles';
+import React from 'react';
+import {StyleSheet, Text, View} from 'react-native';
+import {rem} from 'rn-units';
+
+export const PrivacyTerms = () => {
+  const handlePress = (url: string) => {
```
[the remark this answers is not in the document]
ice-hector: Fixed

src/screens/AuthFlow/InvalidLink/hooks/useInvalidLink.ts
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {useFocusEffect} from '@react-navigation/native';
+import {AccountActions} from '@store/modules/Account/actions';
+import {useCallback} from 'react';
+import {BackHandler} from 'react-native';
+import {useDispatch} from 'react-redux';
+
+export const useInvalidLink = () => {
```
ice-atlas, line 9: Naming. `useBackHandler`?

src/screens/AuthFlow/SignIn/index.tsx
```diff
@@ -28,35 +26,31 @@ import {rem} from 'rn-units';
...
-    emailSentTimestamp,
   } = useEmailAuth();
   const [activeTab, setActiveTab] = useState<Tab>('email');
 
   return (
-    <KeyboardAvoider
-      keyboardVerticalOffset={
-        -SOCIAL_BUTTONS_HEIGHT - DIVIDER_HEIGHT + rem(15)
-      }>
+    <KeyboardAvoider keyboardVerticalOffset={-SOCIAL_BUTTONS_HEIGHT + rem(15)}>
```
ice-atlas, line 47: `rem(15) - SOCIAL_BUTTONS_HEIGHT`

ice-atlas (changes requested):

src/navigation/Router.tsx
```diff
@@ -1,49 +1,51 @@
...
     'iceBonus',
-    'onboarding',
   ];
-  const finalizedAuthSteps =
-    user?.clientData?.registrationProcessFinalizedSteps ?? [];
 
-  if (!isAppInitialized) {
-    return <Initialization />;
+  if (user?.email && user.email.length > 0) {
+    requiredAuthSteps = requiredAuthSteps.filter(step => step !== 'email');
```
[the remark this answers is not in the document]
ice-hector: Removed

src/navigation/Router.tsx
```diff
@@ -1,49 +1,51 @@
...
+  const finalizedAuthSteps =
+    user?.clientData?.registrationProcessFinalizedSteps ?? [];
+
   if (!user) {
     return <AuthNavigator />;
   }
 
-  if (difference(requiredAuthSteps, finalizedAuthSteps).length !== 0) {
+  if (
+    difference(requiredAuthSteps, finalizedAuthSteps).length !== 0 ||
```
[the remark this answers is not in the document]
ice-hector: Refactored

ice-atlas (dismissed):

ice-hades (dismissed):

ice-atlas (approved):

ice-orion (approved):

src/translations/locales/en.json
```diff
@@ -124,21 +123,19 @@
...
+    "description": "Enter the nickname of the person who invited you",
     "inputPlaceholder": "Enter nickname",
-    "dontHaveInvitationCode": "Don't have an invitation code? ",
-    "tapHere": " Tap here!",
-    "dontHaveCodeTip": "If you don't have an invitation code we will\nselect a random member as your referral!",
-    "title_multiline": "Who invited\nyou?",
-    "confirm_text": "Are you sure that you would like to set a random member as your referral?\n\nThis cannot be undone!"
+    "dontHaveInvitationCode": "By completing this step, you and your friend will both receive a {{value}}% bonus on the mining rate when you are both mining in the same time.",
+    "confirm_text": "Are you sure that you would like to set a random member as your referral?\n\nThis cannot be undone!",
+    "dontHaveCodeTip": "If you don't have an invitation code we will\nselect a random member as your referral!"
```
ice-ares, line 138: busted!

ice-ares (approved):

PR #103, opened 2023-01-17
Balance history and mining-stacking improvements

ice-orion:
https://www.notion.so/leftclick/Mining-staking-ca0e13ded0e14ac296fd6e51edec2354

https://www.notion.so/leftclick/Balance-history-00507561069f45e58a5b1d1506fe9021


src/constants/links.ts
```diff
@@ -1,4 +1,14 @@
 // SPDX-License-Identifier: BUSL-1.1
 
-export const APP_STORE_LINK = '';
-export const PLAY_STORE_LINK = '';
+import {ENV} from '@constants/env';
+import {t} from '@translations/i18n';
+
+export const LINKS = {
+  APP_STORE: `https://apps.apple.com/app/${ENV.APPSTORE_APP_ID}`,
```
ice-ares, line 7: Will this work for staging?

src/services/audio/index.ts
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+import {logError} from '@services/logging';
+import {Platform} from 'react-native';
+import Sound from 'react-native-sound';
+
+export enum LocalAudio {
+  startMining = 'start_mining.wav',
```
ice-ares, line 8: Bad design. U have a generic service, with specific data

src/translations/translations.json
```diff
@@ -435,17 +434,19 @@
             "time_left": "Time Left",
             "mining_rate": "Mining Rate",
             "appeal": "Boost your mining rate",
-            "benefits_description": "Pre-stake your mining rewards for up to five years and increase your mining rate by up to 500%",
+            "benefits_description": "Pre-stake your mining rewards for up to 5 years and increase your mining rate by up to 250%",
```
ice-ares, line 437: Parameters!! Never add hardcoded values in translations

src/translations/translations.json
```diff
@@ -484,7 +483,10 @@
             "search_for_users": "Search for users"
         },
         "tabbar": {
-            "mining_tooltip": "Tap on the button below to start your first mining session"
+            "mining_inactive_tooltip": "Tap here to start your first mining session",
+            "mining_reset_tooltip": "Press for 3 seconds to reset your mining session",
+            "mining_holiday_reset_tooltip": "You are using a Day Off! Press for 3 seconds to reset your mining session",
```
ice-ares, line 488: Parameters!

src/constants/links.ts
ice-orion: We won't publish the staging app, so on staging it won't find the actual app (but will open the appstore)

src/services/audio/index.ts
ice-orion: done, added `src/assets/audio/index.ts` for actual data

src/translations/translations.json
ice-orion: done, `constatns/stacking.ts`

src/translations/translations.json
ice-orion: done

ice-ares (dismissed):

ice-ares (dismissed):

src/screens/Staking/components/Footer/index.tsx
```diff
@@ -1,47 +1,86 @@
...
+          onPress: () => {
+            navigation.goBack();
+            IS_STAKING_ACTIVE.current = true;
+          },
+        },
+      ],
+    });
+  };
+
+  let termsAgreeText = reactStringReplace(
```
ice-hades, line 49: can you please move it to a function so you don't have to use `let`

src/screens/Staking/components/Intro/index.tsx
```diff
@@ -2,19 +2,31 @@
...
 import {InfoOutlineIcon} from '@svg/InfoOutlineIcon';
 import {t} from '@translations/i18n';
+import {openLinkWithInAppBrowser} from '@utils/device';
 import {font} from '@utils/styles';
 import React, {memo} from 'react';
 import {StyleSheet, Text} from 'react-native';
 import {rem} from 'rn-units';
 
 export const Intro = memo(() => {
+  const onLinkPress = () => {
```
ice-hades, line 20: no need to declare it as an embedded function. it can be declared on a file level

src/screens/Staking/components/Footer/index.tsx
ice-orion: What do you mean?

src/navigation/components/MainTabBar/components/TabBarMiningItem/components/MiningButtonHandlers.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const activateLongPress = () => {
+    longPressTimer = setTimeout(() => {
+      buttonScale.value = withTiming(1, {duration: 300}, () =>
+        runOnJS(onLongPress)(),
+      );
+    }, MINING_LONG_PRESS_ACTIVATION_SEC * 1000);
+  };
+
+  const clearTimer = () => {
+    if (longPressTimer) {
```
ice-hades, line 46: to my understanding `clearTimer` captures value of `longPressTimer` on the moment of its creation. it won't see any changes to it. what can save here is rerendering and recreation of `clearTimer` with a new latest value of `longPressTimer`. but again, updates to it from `activateLongPress` should not cause any rerenders so I don't see how it works properly. What do you think guys @ice-hector @ice-atlas

src/screens/Staking/components/Footer/index.tsx
ice-hades: mean to avoid declaring a `let` and reassigning to it better stick to the functional programming. if you move this termsAgreeText logic to a function you won't need to declare a `let` and reassign to it

src/navigation/components/MainTabBar/components/TabBarMiningItem/components/MiningButtonHandlers.tsx
ice-orion: >  it won't see any changes to it
Why not? It works fine, i even logged the timer value to make sure :) So any time you start the longPress and drops it, `longPressTimer` has different value (cuz it refers to a different timer)

src/screens/Staking/components/Footer/index.tsx
ice-orion: can you give an example?
currently it works like this:
```
let result = doStuff(text, propsA)
result = doStuff(result, propsB)
```

How do you suggest to handle this?

src/navigation/components/MainTabBar/components/TabBarMiningItem/components/MiningButtonHandlers.tsx
ice-hades: closures capture current values only. if you want to mutate some value a container is used. Like `const longPressTimerContainer = {current: null};`. But better event to use useRef here to store longPressTimer. as it also will prevent from clearing of `longPressTimer` on rerender.

reassigning variables is on the same level as mutating data objects in terms of possible bugs now or in the future

src/screens/Staking/components/Footer/index.tsx
ice-hades: const result = doStuff(text, propsA)
return doStuff(result, propsB)

ice-hades: you can keep it here. main concern is `let`

src/components/CheckBox/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import {Touchable} from '@components/Touchable';
+import {MIDDLE_BUTTON_HIT_SLOP} from '@constants/styles';
+import {CheckboxActive} from '@svg/CheckboxActive';
+import {CheckboxInactive} from '@svg/CheckboxInactive';
+import React from 'react';
+import {StyleProp, ViewStyle} from 'react-native';
+
+type Props = {
+  value: boolean;
```
ice-hector, line 11: value? Rename please to `isActive`

src/navigation/components/MainTabBar/components/TabBarMiningItem/components/MiningButtonHandlers.tsx
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+  const singleTap = Gesture.Tap().onEnd((_event, success) => {
+    if (success) {
+      runOnJS(onTap)();
+    }
+  });
+
+  return (
+    <GestureDetector gesture={longPressActivation ? longTap : singleTap}>
```
ice-hector, line 69: isLongPressActivated

src/navigation/components/MainTabBar/components/TabBarMiningItem/components/MiningButtonHandlers.tsx
ice-orion: > closures capture current values only.

Here is what happens there

image

ice-orion: So closures use a memory reference to a variable, so you always have the correct value

ice-orion: You're probably confusing it with passing a primitive as an arg?

image

ice-hades: you are right, it works

src/components/CheckBox/index.tsx
ice-orion: I took it from the RN naming of the similar component - https://reactnative.dev/docs/switch#value
We also have a custom `Switch` component and we're using name `value` for this prop here and there - i would say it's pretty common

src/utils/device.ts
```diff
@@ -24,3 +29,34 @@ export const openSMS = async (phoneNumber: string, message: string) => {
...
 };
+
+export const openLinkWithInAppBrowser = ({
+  url,
+  options,
+}: {
+  url: string;
+  options?: InAppBrowserOptions;
+}) => {
+  return InAppBrowser.open(url, {
```
ice-hector, line 40: How does it handle the navigation stacks? Or it just opens over everything?

src/screens/Staking/components/Footer/index.tsx
ice-orion: done

src/screens/Staking/components/Intro/index.tsx
ice-orion: done

src/utils/device.ts
ice-orion: yea, it works like a modal or something, so it has nothing to do with our main navigation

src/navigation/components/MainTabBar/components/TabBarMiningItem/components/MiningButtonHandlers.tsx
ice-orion: I would suggest you to reconsider using `is` as prefix for boolean values everywhere (Hungarian notation that is pretty much dead). It is redundant when type-checking is done by the compiler + nowhere in react-native this notation is used (e.g. take a look at boolean props from FlatList - `refreshing`, `inverted`, boolean props from reanimated `AnimationCallback->finished`, react-navigation `LinkingOptions->enabled` and so on).

Sometimes i also add this prefix for const variables for some reason :) But my point is that it is not something required at all...

ice-ares (approved):

src/navigation/components/MainTabBar/components/TabBarMiningItem/components/MiningButtonHandlers.tsx
ice-hector: Yeah, you are right, but refreshing is a process, inverted means should it be done or not (don't know how to explain it correctly). In your case you use `activation` it's a simple noun that does not reflect the meaning of what it does.
Would it be better to add suffix -ed to activated like `longPressActivated`? 
`-ed is added to nouns to form adjectives that describe someone or something as having a particular feature or features`

src/components/CheckBox/index.tsx
ice-hector: Yeah, I also agree but in the scope of CheckBox as I see from the context it is Active or Not active (and related to design). In this case when we name it `value` it is not at all clear what the meaning of this `value` inside the CheckBox (without reading the implementation inside)

ice-hades (dismissed):

src/navigation/components/MainTabBar/components/TabBarMiningItem/components/MiningButtonHandlers.tsx
ice-orion: Well, as for this particular case - this variable belongs to some mining stage. It describes how this particular step might be activated (meaning this boolean describes a step rather than keeping that something  is happened or not).
If it would have been for keeping that activation is happened or not, then yea, it had been fine to name it  `longPressActivated `

src/components/CheckBox/index.tsx
ice-orion: Would it be fine if we named it `checked` (like in react-native-elements)?

ice-ares (approved):

ice-hades (approved):

src/translations/translations.json
```diff
@@ -727,6 +721,12 @@
                     "message": "Are you sure you want to end the {{providerName}} login session from this device?"
                 }
             }
+        },
+        "links": {
+            "pre_staking": "https://ice.io/pre-staking/",
```
ice-atlas, line 726: Do we really want to have links in translations?

ice-atlas (changes requested):

src/assets/svg/SettingsIcons.tsx
```diff
@@ -3,13 +3,14 @@
 import {COLORS} from '@constants/colors';
 import * as React from 'react';
 import Svg, {Path} from 'react-native-svg';
+import {rem} from 'rn-units';
 
 type Props = {
   fill?: string;
 };
 
 export const SettingsIcon = ({fill = COLORS.white}: Props) => (
```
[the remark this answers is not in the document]
ice-orion: fixed, will be in the next pr

src/constants/links.ts
```diff
@@ -1,4 +1,14 @@
...
+  PLAY_STORE: `https://play.google.com/store/apps/details?id=${ENV.APP_ID}`,
+  PRE_STAKING: t('links.pre_staking'),
+  PRIVACY: t('links.privacy'),
```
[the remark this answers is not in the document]
ice-orion: this is a topic to discuss

src/constants/staking.ts
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: BUSL-1.1
+
```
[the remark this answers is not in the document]
ice-orion: this is a topic to discuss

src/translations/translations.json
ice-orion: this is a topic to discuss

src/constants/staking.ts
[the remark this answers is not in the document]
ice-ares: We’ll bother with that later, but yea, we’ll have everything fetched from BE

PR #108, opened 2023-01-24
Feature/linter rules

ice-hector:


.husky/pre-commit
```diff
@@ -5,3 +5,4 @@ npm run check-license
 npm run generate-translation-types
 npm run lint
 npm run check-types
+npx pretty-quick --staged
```
ice-orion, line 8: Why npx? npx is for using packages without installing them as part of the main package.
Since you're specifying it as a dependency, add a script to package.json (e.g. `pretty-staged`) and use `yarn pretty-staged` here

ice-orion (dismissed):

PR #110, opened 2023-01-24
split translations by locale

ice-orion:


scripts/generateTranslationTypes/index.mjs
```diff
@@ -11,9 +11,8 @@ if (!translationsPath) {
 }
 
 const translations = JSON.parse(fs.readFileSync(translationsPath));
-const enTranslations = translations.en;
```
ice-ares, line 14: u still need to default to en

ice-hades (approved):

src/translations/i18n.ts
```diff
@@ -1,10 +1,16 @@
...
-import locales, {Translations} from '@translations/translations.json';
+import en, {Translations} from '@translations/locales/en.json';
+import ro from '@translations/locales/ro.json';
 // eslint-disable-next-line no-restricted-imports
 import {I18n, TranslateOptions} from 'i18n-js';
 import RNLocalize from 'react-native-localize';
 
+const locales = {
+  en,
+  ro,
```
ice-ares, line 11: u sure this is ok? we're gonna import 50 files

scripts/generateTranslationTypes/index.mjs
ice-orion: This is the locale that is going to be used to generate translations types.
And we now pass the en translation as argument (previously we passed the whole translation file and we had to pick en translation from it)

src/translations/i18n.ts
ice-orion: You mean that it will be ugly having all the locales defined here manually? I don't think there is any other way to get the locales from a folder or something like that.
We can probably improve it later, implementing loading this locales on demand (using require)

ice-ares (approved):

PR #123, opened 2023-02-06
chore: phone verification screen

ice-hades:


src/components/Forms/ModifyPhoneNumberForm/hooks/useModifyPhoneNumber.tsx
```diff
@@ -27,13 +40,18 @@ export const useModifyPhoneNumber = () => {
   const modifyPhoneNumber = () =>
     dispatch(
       AccountActions.UPDATE_ACCOUNT.START.create({
-        phoneNumber: fullPhoneRef.current,
+        phoneNumber:
+          fullPhoneRef.current ||
```
ice-ares, line 44: call e164NumberFormat

src/screens/Team/components/Contacts/index.tsx
```diff
@@ -28,24 +29,46 @@ export const Contacts = ({
     isPermissionGrantedSelector('contacts'),
   );
   const isPhoneNumberVerified = useSelector(isPhoneNumberVerifiedSelector);
-  const phoneVerificationStep = useSelector(phoneVerificationStepSelector);
```
ice-orion, line 31: What is this task about? Why we can't rely on tempPhoneNumber now and need those 3 states here?

ice-hades: Go back from confirm code screen and modify the phone (should be filled)

ice-orion: Can't we just add `phoneVerificationStep` to the store and set values there (`code` or `phone`) explicitly?
So to go back we just set `phone` and that's it?
Then we probably won't need to handle it here via states / add all those new props.
Also we have the same step in settings, is it handled there as well?

ice-hades: On settings it should be empty. Confirmed with business team.

Do you see how it can be improved here or just asking why it was done that way and not the way you see it?

ice-orion: I'm just curious if we can do it like this:
keep `temporaryPhoneNumber: {country: string; body: string}`
+ add `phoneVerificationStep: 'code' | 'phone'` to the store

then this file remains the same -> we're showing `ModifyPhoneNumber` / `ConfirmPhoneNumber` relying on `phoneVerificationStep` + we prefill the phone number input with `temporaryPhoneNumber` if it's set (contains both country and body separately)

will it work? 🤔
then we'll rely only on things from store + won't need to handle it manually with states, props passing here and on the settings screen if we even need it there

src/components/Forms/ConfirmPhoneNumberForm/hooks/useConfirmPhoneNumber.ts
```diff
@@ -55,7 +59,7 @@ export const useConfirmPhoneNumber = () => {
 
   const {code, setCode, validationError} = useCodeInput({
     validate,
-    resetValidation,
+    resetValidation: onModifyPhoneNumber ?? resetValidation,
```
ice-hector, line 62: How resetValidation can be onModifyPhoneNumber ? Not clear at all

src/components/Forms/components/ConfirmCode/index.tsx
```diff
@@ -32,6 +39,9 @@ export const ConfirmCode = memo(
         <View style={styles.codeInput}>{CodeInput}</View>
         <View style={styles.resendButton}>{ResendButton}</View>
         <View style={styles.backButton}>{BackButton}</View>
+        {DoThisLaterButton ? (
+          <View style={styles.backButton}>{DoThisLaterButton}</View>
```
ice-hector, line 43: Does it have the same style as the back button?

src/utils/phoneNumber.ts
```diff
@@ -1,12 +1,27 @@
...
+import {getCountryByCode} from '@utils/country';
 import {
   CountryCode,
   formatIncompletePhoneNumber,
   parsePhoneNumberWithError,
 } from 'libphonenumber-js/min';
+import {PhoneNumber} from 'libphonenumber-js/types';
 import {sha256} from 'react-native-sha256';
 
+export function getCountryByPhoneNumber(phoneNumber: string | null): {
```
ice-ares, line 13: what are u doing with this?

ice-hades: parse phone number to get country code and national number

ice-orion (dismissed):

ice-hector (dismissed):

src/components/Forms/ConfirmPhoneNumberForm/hooks/useConfirmPhoneNumber.ts
```diff
@@ -15,7 +15,12 @@ import {
 import {useCallback, useEffect} from 'react';
 import {useDispatch, useSelector} from 'react-redux';
 
-export const useConfirmPhoneNumber = () => {
+type Props = {
+  onModifyPhoneNumber?: () => void;
+};
+
+export const useConfirmPhoneNumber = ({onModifyPhoneNumber}: Props) => {
+  const doNoResetValidationOnGoBack = !!onModifyPhoneNumber;
```
ice-atlas, line 23: `const resetValidationOnGoBack = !onModifyPhoneNumber;`
then
```
if (resetValidationOnGoBack) {
      resetValidation();
    }
```
and adjust `resetValidation`

motivation: easier to read expression with out `!`

src/store/modules/Validation/actions/index.ts
```diff
@@ -1,8 +1,18 @@
 // SPDX-License-Identifier: BUSL-1.1
 
 import {User} from '@api/user/types';
+import {TemporaryPhoneVerificationStepType} from '@store/modules/Validation/reducer';
 import {createAction} from '@store/utils/actions/createAction';
 
+const SET_TEMPORARY_PHONE_VERIFICATION_STEP = createAction(
+  'SET_TEMPORARY_PHONE_VERIFICATION_STEP',
+  {
+    STATE: (
```
ice-atlas, line 10: it is better to use 
```
STATE: (payload: {
  temporaryPhoneVerificationStep: TemporaryPhoneVerificationStepType,
}) => payload
```
easier to extend, less code AND named params for function

ice-atlas (changes requested):

ice-atlas (approved):

ice-orion (dismissed):

ice-atlas (approved):

ice-hector (approved):

PR #130, opened 2023-02-07
Home fixes

ice-atlas:


ice-ares (dismissed):

ice-ares (dismissed):

src/screens/Modals/Tooltip/index.tsx
```diff
@@ -32,6 +36,13 @@ export const Tooltip = () => {
     height: number;
   } | null>(null);
 
+  const [containerHeight, setContainerHeight] = useState(0);
+
+  const onLayout = (event: LayoutChangeEvent) => {
```
ice-orion, line 41: This is basically the whole viewable port height, can't we get it without measuring here?

src/screens/HomeFlow/Home/components/Header/components/MenuButton.tsx
```diff
@@ -38,11 +37,15 @@ export const MenuButton = memo(() => {
           right: screenWidth - x - rem(16),
         },
         buttons: [
-          {
-            icon: <BellIcon color={COLORS.downriver} />,
-            label: t('home.menu.notifications'),
-            onPress: () => navigation.navigate('Notifications'),
-          },
+          // Hide until functionality is ready
```
ice-orion, line 40: Probably it would be better to mark these kind of places with `TODO:` ? This way we would know the places in the code that require our attention and improvement?

src/screens/HomeFlow/Home/components/Overview/components/OnlineUsersHistory/index.tsx
```diff
@@ -53,17 +53,20 @@ export const OnlineUsersHistory = () => {
...
-                valuePercentage={valuePercentage}
-                label={item.label}
-              />
-            </View>
-          );
-        })}
+        {data
+          .slice()
+          .reverse()
+          .map(item => {
```
ice-orion, line 59: optional: just add `reverse` after `map`

src/screens/Modals/Tooltip/index.tsx
ice-atlas: I am open for propositions. But this is the only one solid solution for this. Previous implementation was not accurate for some devices

ice-orion (approved):

ice-hector (approved):

ice-ares (dismissed):

ice-hector (dismissed):

ice-hector (approved):

ice-ares (approved):

ice-ares:
@ice-atlas fix whatever else u need to fix in a new PR

PR #155, opened 2023-02-20
disable profile

ice-orion:


ice-ares (approved):

src/components/ListItems/UserListItem/index.tsx
```diff
@@ -46,7 +46,8 @@ export const UserListItem = memo(
       <View style={styles.container} {...stopPropagation}>
         <Touchable
           style={styles.touchArea}
-          onPress={() => navigation.navigate('UserProfile', {userId})}>
+          // TODO: temp profile disabling
+          onPress={() => false && navigation.navigate('UserProfile', {userId})}>
```
ice-atlas, line 50: I may suggest just to comment UserProfile from Navigator. All the places will silently fail to navigate

ice-hector (approved):

ice-atlas (changes requested):

src/components/ListItems/UserListItem/index.tsx
ice-orion: Why is it better? We'll get errors in sentry...

ice-atlas: Oh, I see. Then OK

ice-atlas (approved):

PR #178, opened 2023-03-06
chore: add set correct firebase locale on app launch

ice-hector:


ice-atlas (approved):

src/store/modules/Account/sagas/syncLanguageCode.ts
```diff
@@ -13,6 +13,7 @@ import {call, SagaReturnType, select} from 'redux-saga/effects';
  * Check user.language property and react on locale change
  */
 export function* syncLanguageCodeSaga() {
+  yield call(setAuthLanguageCode, getLocale());
```
ice-ares, line 16: what about the one from user?

ice-orion: may be it would be better to set
```
setDayjsLocale(getLocale());
```
also here instead of `services/dayjs`? So we keep everything in 1 place

ice-orion: @ice-ares it is handle later on in this file:
we're listening to the user.language and if user.language is not equal to `getLocale()`
we're calling `setLocale` + `setDayjsLocale` + `setAuthLanguageCode`

PR #180, opened 2023-03-14
Feature/locales

ice-aiolos:


ice-ares (dismissed):

ice-atlas (approved):

ice-atlas (approved):

src/services/dayjs/index.ts
ice-orion, line 1: Not directly a part of the pr but probably it would be better to call `setDayjsLocale` in `syncLanguageCodeSaga`?
To keep all the locale related calls in 1 place?

ice-orion (approved):

src/services/dayjs/index.ts
ice-aiolos: let's sync tomorrow and discuss

